PR #1, opened 2016-11-01
Native, native swift and react native templates for iOS

wmathurin:
They all use npm + pod to get mobile sdk (and other dependencies)
They all have a template.js script which is invoked by the new forceios to inject app name / company id etc.
The template app can be run directly (after first running install.sh or installios.sh in the case of the react native template).
The react native template is for both ios and android and was generated with react-native cli and is compatible with react-native cli.


wmathurin:
The android code in the react native template has not yet been tested.

wmathurin:
The public->private change unlinked this repo from my fork.
Closing the pull request. Will re-open once I get the fork back.

PR #11, opened 2016-12-01
add build status on README page

helenren:


wmathurin:
Once we ship 5.0, should we have a different link in master and unstable?

PR #12, opened 2016-12-06
overlay in dev mode fix

ivanbogdanov:


wmathurin:
Will you also fix SmartSyncExplorerReactNative?

ReactNativeTemplate/android/app/build.gradle
```diff
@@ -109,6 +109,11 @@ android {
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
         }
+        debug {
+            defaultConfig {
+                targetSdkVersion 22
```
bhariharan: @ivanbogdanov We want this to be `25` as well. Setting this to `22` is not solving the problem, it's simply supporting a version before the new security popup feature was introduced in `Android`. We need the `targetSdkVersion` to be the same for both `debug` and `release`. I would revert the changes to this file. We need to fix this while keeping `targetSdkVersion` as `25`.

ivanbogdanov: Yes it is dev only" but the initial crush look really ugly And it actually cover several other issues that were revealed later. 
For unexperienced developers  is can be very confusing experience.

I can look into ReactActivity fix but it may take months before it goes thru

bhariharan: We've had this for 2 releases now and no one has complained yet. I would prefer to leave it as is until we determine a proper fix. If someone complains or reaches out, we can ask them to drop the API version.

wmathurin: I'm okay with using RN target sdk version. Our SDK will work with 22, we don't know if theirs will work with 25. So it makes sense to go with the one we know will work for both.

ivanbogdanov: OK. Let me quickly look into one more option. If it works out we may keep 25

ivanbogdanov: Cool:  I am planning to create one more PR to SalesforceReactActivity to address it. If this PR is approved we could keep targetSdkVersion 25 here

bhariharan: If there's a fix in `SalesforceReactActivity` that lets us keep `targetApiVersion = 25`, I would prefer that.

bhariharan:
@ivanbogdanov With [this](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1319) change, can this PR be closed?

ivanbogdanov:
@bhariharan strings.xml change is still good in this PR: I will remove build.gradle change from this PR

PR #25, opened 2017-03-01
Fixing package test failures

bhariharan:


AndroidNativeTemplate/build.gradle
```diff
@@ -4,10 +4,15 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:2.2.2'
+    classpath 'com.android.tools.build:gradle:2.2.3'
+    classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
```
bhariharan: The additional plugins we added need to be in the classpath for the template.

bhariharan:
I'm looking into `hybrid`.

PR #26, opened 2017-03-01
Classpath now comes from the libraries themselves

bhariharan:


bhariharan:
We don't need the inclusions in the top level `build.gradle` anymore, since it's compartmentalized to the library project's `build.gradle`.

PR #32, opened 2017-04-11
Fixing double login bug on iOS

bhariharan:


ReactNativeTemplate/index.ios.js
```diff
@@ -48,7 +48,7 @@ var App = React.createClass({
 
     componentDidMount: function() {
         var that = this;
-        oauth.authenticate(
+        oauth.getAuthCredentials(
```
bhariharan: `authenticate` on `iOS` always brings up the login screen. We need `getAuthCredentials`, which brings up the login screen only if required.

ReactNativeTemplate/ios/ReactNativeTemplate/AppDelegate.m
```diff
@@ -52,7 +52,7 @@ - (id)init
         [SalesforceSDKManager sharedManager].connectedAppCallbackUri = OAuthRedirectURI;
         [SalesforceSDKManager sharedManager].authScopes = @[ @"web", @"api" ];
         // Uncomment the following line if you don't want login to happen when the application launches
-        // [SalesforceSDKManager sharedManager].authenticateAtLaunch = NO;
+        [SalesforceSDKManager sharedManager].authenticateAtLaunch = NO;
```
bhariharan: Let the plugin handle authentication.

PR #41, opened 2017-06-29
Speeding up SDK dependencies download

wmathurin:
Using a shallow git clone instead of npm dependencies (which were also getting submodules)

### DIFF
--- a/ReactNativeTemplate/ios/Podfile
+++ b/ReactNativeTemplate/ios/Podfile
@@ -1,4 +1,4 @@
-platform :ios, '10.0'
+platform :ios, '9.0'
 
 project 'ReactNativeTemplate.xcodeproj'
 target 'ReactNativeTemplate' do
--- a/ReactNativeTemplate/ios/Podfile
+++ b/ReactNativeTemplate/ios/Podfile
@@ -1,5 +1,4 @@
-# Uncomment this line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '10.0'
 
 project 'ReactNativeTemplate.xcodeproj'
 target 'ReactNativeTemplate' do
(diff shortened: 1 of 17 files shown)


ReactNativeTemplate/ios/Podfile
```diff
@@ -1,5 +1,4 @@
-# Uncomment this line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '10.0'
```
bhariharan: Shouldn't this be `9.0`? That's the minimum we support, not `10.0`.

iOSNativeSwiftTemplate/Podfile
```diff
@@ -1,5 +1,4 @@
-# Uncomment this line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '10.0'
```
bhariharan: Should be `9.0`.

iOSNativeTemplate/Podfile
```diff
@@ -1,5 +1,4 @@
-# Uncomment this line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '10.0'
```
bhariharan: Should be `9.0`.

iOSNativeTemplate/install.js
```diff
@@ -1,7 +1,20 @@
...
 var execSync = require('child_process').execSync;
-console.log('Installing npm dependencies');
-execSync('npm install', {stdio:[0,1,2]});
+var path = require('path');
+var fs = require('fs');
+
+console.log('Installing sdk dependencies');
+for (var sdkDependency in packageJson.sdkDependencies) {
+    var repoUrlWithBranch = packageJson.sdkDependencies[sdkDependency];
+    var parts = repoUrlWithBranch.split('#'), repoUrl = parts[0], branch = parts.length > 1 ? parts[1] : 'master';
```
bhariharan: Why `master`? Shouldn't we use tag? What exactly is this doing?

wmathurin: I read the SDK repo URL from the package.json. Then I do a git clone (shallow). But first I need to extract the tag or branch from the URL. Repo#branch is a syntax understood by npm not git. If the URL doesn't have a tag or branch specified then it means one wants master. I could also drop the branch option from the git clone command in that case.

ReactNativeTemplate/ios/Podfile
wmathurin: I will do that - devs could change the podfile and rerun pod update it they want 9.0.

wmathurin:
Target iOS 9 by default: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/41/commits/93c3eba410ac16b119f5d3cdd1235925cfcdea6b

iOSNativeTemplate/install.js
bhariharan: Sounds good.

wmathurin:
Unfortunately we need to get the submodules: some of them (e.g. fmdb) are needed to compile the sdk - some are useless for native/react native apps (e.g. cordova, react-native, shared).
I'll send a commit with the change to the install.js that need it.

wmathurin:
The connection can get slow here. It makes a big difference. Not having to download the cordova/react-native submodules is a time saver.

PR #46, opened 2017-07-07
Fixes for react native templates

wmathurin:
See commits for details


wmathurin:
We don't use whatwg-fetch and in react-native cli generated app, it's not included by default.

PR #50, opened 2017-07-14
FMDB is now coming from cocoapods

wmathurin:
However because of the XCode 9 issue, getting it from branch 2.7.2_xcode9 on our fork (which has 2.7.2 + import fix for xcode 9)

### DIFF
--- a/ReactNativeTemplate/ios/Podfile
+++ b/ReactNativeTemplate/ios/Podfile
@@ -27,6 +27,9 @@ pod 'React', :path => '../node_modules/react-native', :subspecs => [
     'RCTLinkingIOS'
 ]
 
+# NB: get rid of next line once FMDB works with xcode 9
+pod 'FMDB', :git => 'https://github.com/forcedotcom/fmdb', :branch => '2.7.2_xcode9' 
+
 pod 'SalesforceAnalytics', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SalesforceSDKCore', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SmartStore', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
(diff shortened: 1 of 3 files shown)


wmathurin:
Also see: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2136

ReactNativeTemplate/ios/Podfile
```diff
@@ -27,6 +27,9 @@ pod 'React', :path => '../node_modules/react-native', :subspecs => [
     'RCTLinkingIOS'
 ]
 
+# NB: get rid of next line once FMDB works with xcode 9
+pod 'FMDB', :git => 'https://github.com/forcedotcom/fmdb', :branch => '2.7.2_xcode9'
```
wmathurin: These will go away once FMDB does the right thing.

bhariharan: So, once `FMDB` merges, we'll go back to inline compiling?

wmathurin: No actually once FMDB merges, we can drop that line from our pod files
```
pod 'FMDB', :git => 'https://github.com/forcedotcom/fmdb', :branch => '2.7.2_xcode9'
```

And we will use the office FMDB pod instead because of that [line](https://github.com/wmathurin/SalesforceMobileSDK-iOS/blob/372cba2e5f20835bd4df017bd12ae42c7231e38b/SmartStore.podspec#L23) in SmartStore.podspec

The inline compiling will just be for the main repo.

bhariharan: Gotcha. So we leave it inlined for our workspace but use podspec from vanilla fmdb elsewhere. That seems like the best of both worlds. IIRC the fmdb project file had stuff that was missing which is another reason we compiled inline.

wmathurin: Exactly, so no impact to internal teams or cordova plugin, but for cocoapods consumers of the SDK it will be treated like other dependencies.

PR #51, opened 2017-07-19
No need to get fmdb submodule since we get FMDB through cocoapod

wmathurin:


wmathurin:
It's because of that change: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2136

PR #58, opened 2017-08-12
Changes to hybrid template so they work with npm 5

wmathurin:
We already made those changes to the other templates (in Mobile SDK 5.2).
npm 5 does not download http dependencies, so you end up missing force.js.


wmathurin:
@bhariharan yes.

PR #72, opened 2017-09-21
Fixing templates (after SFRestDelegate change)

wmathurin:


wmathurin:
The problem with the change to SFRestDelegate (addition of raw response), is that there is no complains at compile time or run time, but of course the methods (for the "old" SFRestDelegate) are never invoked so the application misbehaves.

Could we revisit that change to make it less disruptive?
Maybe SFRestDelegate could have the old signature method and the new signature method and the code would invoke the new one if available and otherwise call the old one?

wmathurin:
@bhariharan True. Unfortunately, our templates still use SFRestDelegate. Anyway with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2210/, we can remain backwards compatible.

PR #76, opened 2017-10-31
Adding native Kotlin support and support for Gradle 3.0

bhariharan:


AndroidNativeKotlinTemplate/app/build.gradle
```diff
@@ -1,14 +1,12 @@
 apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
```
bhariharan: We don't need this plugin anymore. It's natively supported in `Android Studio 3.0`.

PR #82, opened 2017-11-09
Update minimum version to iOS 10

brandonpage:


wmathurin:
@brandonpage You also need to update the ios project and podfile in https://github.com/forcedotcom/SmartSyncExplorerReactNative

PR #95, opened 2017-12-09
Swift template now pulling in SalesforceSwiftSDK

wmathurin:
Note: it compiles and runs but it's not yet using the new rest api wrappers

### DIFF
--- a/iOSNativeSwiftTemplate/Podfile
+++ b/iOSNativeSwiftTemplate/Podfile
@@ -11,6 +11,8 @@ pod 'SalesforceAnalytics', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SalesforceSDKCore', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SmartStore', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SmartSync', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
+pod 'SalesforceSwiftSDK', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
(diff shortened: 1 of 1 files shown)


wmathurin:
Requires: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2316

iOSNativeSwiftTemplate/Podfile
```diff
@@ -11,6 +11,8 @@ pod 'SalesforceAnalytics', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SalesforceSDKCore', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SmartStore', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
 pod 'SmartSync', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
+pod 'SalesforceSwiftSDK', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
```
bhariharan: Should [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2316/files#diff-c555baabfe30a4cd5690fe84a35fb24eR23) be `5.0.3` as well?

wmathurin: For the tag I need to specify the exact tag, for the dependency, we don't have too.
Sp maybe [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2316/files#diff-c555baabfe30a4cd5690fe84a35fb24eR23) should be ~> 5.0.0 to match any of the 5.0.x.

wmathurin: I'll make the change and make sure the submodule also points to 5.0.3.

PR #97, opened 2017-12-10
Update React Native Template to use single entry point

brandonpage:


bhariharan:
Consolidate these two in `__tests__` as well -> `https://github.com/forcedotcom/SalesforceMobileSDK-Templates/tree/dev/ReactNativeTemplate/__tests__`.

PR #99, opened 2017-12-14
Changes to reflect new Swift api(s)

trooper2013:


iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/AppDelegate.swift
```diff
@@ -39,40 +39,36 @@ class AppDelegate : UIResponder, UIApplicationDelegate
     init()
     {
         super.init()
-        SalesforceSDKManager.setInstanceClass(SalesforceSwiftSDKManager.self)
-        SalesforceSDKManager.shared().appConfig?.remoteAccessConsumerKey = RemoteAccessConsumerKey
-        SalesforceSDKManager.shared().appConfig?.oauthRedirectURI = OAuthRedirectURI
-        SalesforceSDKManager.shared().appConfig?.oauthScopes = ["web", "api"];
-
+        
         //Uncomment the following line inorder to enable/force the use of advanced authentication flow.
```
bhariharan: Could we add another builder step called `postInit` where we move these commented out blocks of code? This would be before `postLaunch`.

PR #100, opened 2017-12-14
Switching to SmartSyncSDKManager by default

bhariharan:
@wmathurin @rwhitley


wmathurin:
Don't you think developers will swap that activity for theirs? Should we do it in the MainApplication.kt class instead (which is less likely to be touched)?

PR #101, opened 2017-12-15
Fixing bug with assembleRelease

bhariharan:
- We need to explicitly override the value of `entryFile` in `react.gradle` -> it defaults to `index.android.js`, but we use a unified `index.js`.
- Removing the directories with rogue symlinks that break `assembleRelease` with `rimraf`.

### DIFF
--- a/ReactNativeTemplate/installandroid.js
+++ b/ReactNativeTemplate/installandroid.js
@@ -4,17 +4,18 @@ var packageJson = require('./package.json')
 var execSync = require('child_process').execSync;
 var path = require('path');
 var fs = require('fs');
-
+var rimraf = require('rimraf');
 console.log('Installing sdk dependencies');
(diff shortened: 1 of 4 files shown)


ReactNativeTemplate/installandroid.js
```diff
@@ -4,17 +4,18 @@ var packageJson = require('./package.json')
 var execSync = require('child_process').execSync;
 var path = require('path');
 var fs = require('fs');
-
+var rimraf = require('rimraf');
```
wmathurin: The pb is that we don't run npm install until line 21.
So if you start clean (remove node_modules), it will fail.

wmathurin: So we should move up the npm install call or move down the rimraf related stuff (or just use --recurse-submodules in the git call).

PR #109, opened 2018-02-21
Upgrading to ReactNative 0.53.3

bhariharan:


bhariharan:
The `Android` app compiles and runs successfully. I haven't updated the `iOS Podspecs` yet. Will update that and verify.

ReactNativeTemplate/package.json
```diff
@@ -11,10 +11,11 @@
         "SalesforceMobileSDK-iOS": "https://github.com/forcedotcom/SalesforceMobileSDK-iOS.git#dev"
     },
     "dependencies": {
-        "react": "16.0.0",
-        "react-native": "0.50.4",
+        "react": "16.2.0",
```
wmathurin: I wonder if we even need these here since react-native-force should pull in anything it needs.

bhariharan: It complains: `UNMET PEER DEPENDENCY`.

wmathurin: Hmm that sounds familiar - maybe I did try that before.

bhariharan:
The `iOS` app compiles and runs successfully as well.

PR #121, opened 2018-06-06
 File listing templates with description etc

wmathurin:


templates.json
```diff
@@ -0,0 +1,77 @@
+[ 
+    {
+
+        "path": "iOSNativeSwiftTemplate",
+        "description": "Basic Swift application",
```
wmathurin: @rwhitley Could you review the descriptions? They will be shown to the use when using the new forceios/forcedroid/forcereact/forcehybrid listtemplates (or sfdx mobilesdk:ios:listtemplates).

PR #123, opened 2018-06-11
Bumping up Gradle version

bhariharan:


bhariharan:
Good catch. Will add that.

PR #124, opened 2018-06-25
Better usability: delete/undelete button below fields

wmathurin:
Instead of being next to "Save" button.
Now the same as native SmartStoreExplorer apps.


wmathurin:
New look:
screen shot 2018-06-25 at 5 32 22 pm

PR #134, opened 2018-07-06
Using promises in StoreMgr

wmathurin:
* Using promiserNoRejection for sync operations to gracefully handle errors while offline
* Going to the end of the result set when querying smartstore (before we were just getting first 100)

### DIFF
--- a/SmartSyncExplorerReactNative/js/StoreMgr.js
+++ b/SmartSyncExplorerReactNative/js/StoreMgr.js
@@ -25,7 +25,14 @@
  */
 
 import EventEmitter from './events';
-import {smartstore, smartsync} from 'react-native-force';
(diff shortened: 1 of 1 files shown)


SmartSyncExplorerReactNative/js/StoreMgr.js
```diff
@@ -25,7 +25,14 @@
  */
 
 import EventEmitter from './events';
-import {smartstore, smartsync} from 'react-native-force';
+import {smartstore, smartsync, forceUtil} from 'react-native-force';
+
+const registerSoup = forceUtil.promiser(smartstore.registerSoup);
```
wmathurin: Using promise-based version of the APIs
For sync operations using promiserNoRejection which funnels success and error to the `then()`

PR #136, opened 2018-07-09
Showing last error if any

wmathurin:
If record fail to sync, the sync-problem icon is shown in the list and the error message is shown in the detail screen.
Changing the record removes the last error.
screen shot 2018-07-09 at 4 40 50 pm
screen shot 2018-07-09 at 4 40 58 pm

### DIFF
--- a/SmartSyncExplorerReactNative/js/ContactScreen.js
+++ b/SmartSyncExplorerReactNative/js/ContactScreen.js
@@ -26,6 +26,7 @@
 
 import React from 'react';
 import {
+    Platform,
     ScrollView,
     StyleSheet,
     View,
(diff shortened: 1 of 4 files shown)


SmartSyncExplorerReactNative/js/ContactScreen.js
```diff
@@ -61,6 +61,34 @@ const ContactScreen = createReactClass({
         storeMgr.saveContact(contact, () => {this.props.navigator.pop()});
     },
 
+    renderErrorIfAny() {
+        var errorMessage = null;
+        const lastError = this.state.contact.__last_error__;
+        if (lastError) {
+            try {
+                errorMessage = new RegExp('.*message = "([^"]*)".*', 'm').exec(lastError)[1];
```
wmathurin: This is not pretty.

However I just realized, I passed the raw response on [Android](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncUpTarget.java#L204) but not on [iOS](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SmartSync/SmartSync/Classes/Target/SFSyncUpTarget.m#L229). If I were to pass the raw response, we could do a `JSON.parse(lastError).message`.

PR #137, opened 2018-07-11
Not creating sync down/up with every call

wmathurin:
Creating sync down and sync up for each object type at the beginning.
Then simply calling reSync(byName:syncName).
(It will do a full sync the first time and then only delta syncs).

### DIFF
--- a/BaristaSwift/Common/Common/Base/Store.swift
+++ b/BaristaSwift/Common/Common/Base/Store.swift
@@ -36,34 +36,46 @@ public typealias SyncCompletion = ((SFSyncState?) -> Void)?
 public class Store<objectType: StoreProtocol> {
 
     private final let pageSize: UInt = 100
-    
-    public init() {
-//        self.store.removeAllSoups()
-//        store.clearSoup(objectType.objectName)
-//        store.removeSoup(objectType.objectName)
-    }
 
+    private final let syncDownName: String = "syncDown_\(objectType.objectName)";
+    
+    private final let syncUpName: String = "syncUp_\(objectType.objectName)";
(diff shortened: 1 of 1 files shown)


BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -137,26 +149,8 @@ public class Store<objectType: StoreProtocol> {
         }
     }
 
-    public func syncDown<T:StoreProtocol>(child: T, completion: SyncCompletion = nil ) {
```
wmathurin: These methods are never used so getting rid of them

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -167,37 +161,17 @@ public class Store<objectType: StoreProtocol> {
...
         
         DispatchQueue.main.async(execute: {
-            let options: SFSyncOptions = SFSyncOptions.newSyncOptions(forSyncUp: objectType.readFields, mergeMode: .overwrite)
-            let target = SFSyncUpTarget.init(createFieldlist: objectType.createFields, updateFieldlist: objectType.updateFields)
-            self.smartSync.syncUp(with: target, options: options, soupName: objectType.objectName, update: updateBlock)
+            self.smartSync.reSync(byName: self.syncUpName, update: updateBlock)
         })
     }
     
-    public func syncUpDownResolvingChildren<P:StoreProtocol, C:StoreProtocol>(parent:P, child: C, completion: SyncCompletion = nil) {
```
wmathurin: This method is never used.

PR #138, opened 2018-07-13
Using promises when synching

wmathurin:
Refactored Store.swift (base class for all Store objects) to use promises when synching.
As a result, had to refactor the callers (and their callers etc) as well.
That caused error handling to be simplified (if any error happens, the promise is "failed" - so in case of success the promise resolves with the expected object).

See comments in PR for more details.

### DIFF
--- a/BaristaSwift/Common/Common/Base/Store.swift
+++ b/BaristaSwift/Common/Common/Base/Store.swift
@@ -30,8 +30,7 @@ import SalesforceSDKCore
 import SalesforceSwiftSDK
 import SmartStore
 import SmartSync
-
-public typealias SyncCompletion = ((SFSyncState?) -> Void)?
+import PromiseKit
 
 public class Store<objectType: StoreProtocol> {
 
@@ -54,7 +53,7 @@ public class Store<objectType: StoreProtocol> {
             do {
                 try store.registerSoup(soupName, withIndexSpecs: indexSpecs, error: ())
             } catch let error as NSError {
-                SalesforceSwiftLogger.log(type(of:self), level:.error, message: "\(soupName) failed to register soup: \(error.localizedDescription)")
+                SalesforceSwiftLogger.log(type(of:self), level:.error, message: "\(objectType.objectName) failed to register soup: \(error.localizedDescription)")
             }
         }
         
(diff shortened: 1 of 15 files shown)


BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -30,8 +30,7 @@ import SalesforceSDKCore
 import SalesforceSwiftSDK
 import SmartStore
 import SmartSync
-
-public typealias SyncCompletion = ((SFSyncState?) -> Void)?
```
wmathurin: Every function that had that completion block now returns a Promise<Void> instead.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -54,7 +53,7 @@ public class Store<objectType: StoreProtocol> {
             do {
                 try store.registerSoup(soupName, withIndexSpecs: indexSpecs, error: ())
             } catch let error as NSError {
-                SalesforceSwiftLogger.log(type(of:self), level:.error, message: "\(soupName) failed to register soup: \(error.localizedDescription)")
+                SalesforceSwiftLogger.log(type(of:self), level:.error, message: "\(objectType.objectName) failed to register soup: \(error.localizedDescription)")
```
wmathurin: We happen to name the soup after the object type's name but I think this code is more readable that way.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
         }
         return results
     }
-
-    public func upsertEntries(jsonResponse: Any, completion: SyncCompletion = nil) {
```
wmathurin: Function was never called.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
-    }
-    
-    public func upsertEntries<T:StoreProtocol>(record: T, completion: SyncCompletion = nil) {
-        store.upsertEntries([record.data], toSoup: T.objectName)
-        completion?(nil)
-    }
     
-    public func upsertNewEntries<T:StoreProtocol>(entry: T, completion: SyncCompletion = nil) {
-        var record: T = entry
+    public func locallyCreateEntry(entry: objectType) -> objectType {
```
wmathurin: Better name for the functions.
Now we have three methods that only do local changes: locallyCreateEntry, locallyUpdateEntry and locallyDeleteEntry.
And also three methods that do the local changes followed by a sync: createEntry, updateEntry and deleteEntry.
Also it didn't make sense to have generic methods since the whole class is a generic type.
Finally the create and update methods are returning the object they created or updated (that simplifies calling code).

-

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
+    }
+
+    public func createEntry(entry: objectType) -> Promise<objectType> {
+        _ = locallyCreateEntry(entry: entry)
+        return syncUpDown()
+            .then { _ -> Promise<objectType> in
+                if let record = self.record(forExternalId: entry.externalId) {
+                    return Promise.value(record)
+                } else {
+                    return Promise(error:StoreErrors.recordNotFound)
```
wmathurin: Thanks to promises, the code gets cleaner: instead of dealing with an optional that the caller would have to guard against, the caller is guaranteed to get the object in the then block. In case of error, the catch block of the promise chain will be invoked.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
-                if syncState.isDone() || syncState.hasFailed() {
-                    DispatchQueue.main.async {
-                        if syncState.hasFailed() {
-                            SalesforceSwiftLogger.log(type(of:self), level:.error, message:"syncUp \(objectType.objectName) failed")
-                        }
-                    }
-                    completion?(syncState)
+    private func reSync(syncName: String) -> Promise<Void> {
+        return smartSync.Promises.reSync(syncName: syncName)
+            .then { syncState -> Promise<Void> in
```
wmathurin: @trooper2013 I think we should change our SmartSync extensions to have the sync promises resolve when the sync is done and reject when the sync has failed. That way, I wouldn't to do what I did here. Getting a failed sync state in the next then complicates the code of the callers.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
+    public func syncDown() -> Promise<Void> {
+        return reSync(syncName: syncDownName)
+    }
+    
+    public func syncUp() -> Promise<Void> {
+        return reSync(syncName: syncUpName)
+    }
+    
+    public func syncUpDown() -> Promise<Void> {
+        return self.syncUp().then { self.syncDown() }
```
wmathurin: Almost as concise as JavaScript!

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -213,4 +204,10 @@ public class Store<objectType: StoreProtocol> {
         }
         return objectType.from(results)
     }
+    
+    enum StoreErrors : Error {
```
wmathurin: I have errors in multiple classes, maybe I should centralize them?

BaristaSwift/Common/Common/Store/AccountStore.swift
```diff
@@ -41,6 +42,24 @@ public class AccountStore: Store<Account> {
         return self.account(userId)
     }
     
+    public func getOrCreateMyAccount() -> Promise<Account> {
```
wmathurin: Both apps' AppDelegate had code doing basically what this method does. So I moved it here (and modified to use promises).

BaristaSwift/Common/Common/Store/FavoritesStore.swift
```diff
@@ -47,11 +48,17 @@ public class FavoritesStore: Store<UserFavorite> {
         return UserFavorite.from(results)
     }
     
-    public func addNewFavorite(_ forProduct:Product, completion:SyncCompletion) {
-        guard let account = AccountStore.instance.myAccount(), let accountId = account.accountId else { return }
+    public func addNewFavorite(_ forProduct:Product) -> Promise<UserFavorite> {
+        guard let account = AccountStore.instance.myAccount(), let accountId = account.accountId else {
```
wmathurin: Again, notice the error handling, if a guard fails, I fail the promise, so the code in the then blocks only deals with the normal case.

BaristaSwift/Common/Common/Store/LocalCartStore.swift
```diff
@@ -105,32 +106,34 @@ public class LocalCartStore {
             switch optionType {
             case .integer:
                 if withOption.quantity > 0 {
-                    print("updating quantity for: \(self.inProgressItem!.options[index].product.productName) to: \(withOption.quantity)")
+                    print("updating quantity for: \(String(describing: self.inProgressItem!.options[index].product.productName)) to: \(withOption.quantity)")
```
wmathurin: Fixing a warning.

BaristaSwift/Common/Common/Store/LocalCartStore.swift
```diff
@@ -205,131 +184,105 @@ public class LocalCartStore {
...
-                if let complete = syncState?.isDone(), complete == true {
-                    SalesforceSwiftLogger.log(type(of:self), level:.info, message:"create new quote - sync completed")
-                    guard let synced = QuoteStore.instance.record(forExternalId: newQuoteId) else {
-                        completion(nil)
-                        return
-                    }
-                    opportunity.primaryQuote = synced.quoteId
-                    OpportunityStore.instance.upsertEntries(record: opportunity)
-                    OpportunityStore.instance.syncUp()
-                    completion(synced)
```
wmathurin: The old code would fire the completing code without waiting for the opportunity sync up.
I changed that for now, but it might look slower as a result.

BaristaSwift/Common/Common/Store/LocalOrderStore.swift
```diff
@@ -73,31 +74,19 @@ public class LocalOrderStore {
...
-                        completion(true)
-                    } else if optyState.hasFailed() {
-                        SalesforceSwiftLogger.log(type(of:self), level:.info, message:"Failed syncing opportunity update")
-                        completion(false)
-                    }
-                })
-            } else if quoteState.hasFailed() {
-                SalesforceSwiftLogger.log(type(of:self), level:.info, message:"Failed syncing quote update")
-                completion(false)
+        return QuoteStore.instance.updateEntry(entry: quote)
```
wmathurin: Notice how much shorter the code is because I don't have to deal with the sync state and also I get the updated entry back from updateEntry() (and it's guaranteed to be an object).

BaristaSwift/Common/Common/Store/LocalOrderStore.swift
```diff
@@ -73,31 +74,19 @@ public class LocalOrderStore {
...
+    public func completeOrder(_ order:LocalOrder) -> Promise<Void> {
         let opty = order.opportunity
         opty.stage = .closedWon
         let quote = order.quote
         quote.status = .accepted
         
-        QuoteStore.instance.updateEntry(entry: quote) { (quoteSync) in
-            guard let quoteState = quoteSync else { return }
-            if quoteState.isDone() {
-                SalesforceSwiftLogger.log(type(of:self), level:.info, message:"update quote completed")
```
wmathurin: The base class takes care of doing some logging.

BaristaSwift/Common/Common/Store/LocalOrderStore.swift
```diff
@@ -106,75 +95,26 @@ public class LocalOrderStore {
...
+            QuoteLineGroupStore.instance.syncDown(),
+            OpportunityStore.instance.syncDown()];
         
-        UserStore.instance.syncDown(completion: syncCompletion)
-        ProductOptionStore.instance.syncDown(completion: syncCompletion)
-        QuoteStore.instance.syncDown(completion: syncCompletion)
-        QuoteLineItemStore.instance.syncDown(completion: syncCompletion)
-        QuoteLineGroupStore.instance.syncDown(completion: syncCompletion)
-        OpportunityStore.instance.syncDown(completion: syncCompletion)
+        return when(fulfilled:syncs)
```
wmathurin: Running the sync down in parallel.

wmathurin: See how much more readable the code is.

BaristaSwift/Provider/Provider/View Controller/ViewController.swift
```diff
@@ -140,16 +142,15 @@ class ViewController: UIViewController {
...
-            self.refreshControl.endRefreshing()
-        }
+        _ = self.runSync()
+                .done {
+                    self.updateData()
+                    self.refreshControl.endRefreshing()
+                }
     }
     
-    func runSync(completion:@escaping () -> Void) {
```
wmathurin: runSync() used to call updateData(), I moved the updateData() to the done block of callers of runSync().

BaristaSwift/Common/Common/Base/Store.swift
bhariharan: Sounds like a good idea.

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
-            if let upComplete = upState?.isDone(), upComplete == true {
-                self.syncDown(completion: completion)
-            }
-        }
+    public func syncDown() -> Promise<Void> {
+        return reSync(syncName: syncDownName)
+    }
+    
+    public func syncUp() -> Promise<Void> {
+        return reSync(syncName: syncUpName)
```
bhariharan: Shouldn't this be `syncUp(syncName: syncUpName)`?

BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -89,95 +88,87 @@ public class Store<objectType: StoreProtocol> {
...
     }
     
-    public func syncUpDown(completion: SyncCompletion) {
-        self.syncUp { (upState) in
-            if let upComplete = upState?.isDone(), upComplete == true {
-                self.syncDown(completion: completion)
-            }
-        }
+    public func syncDown() -> Promise<Void> {
+        return reSync(syncName: syncDownName)
```
bhariharan: Shouldn't this be `syncDown(syncName: syncDownName)`?

BaristaSwift/Common/Common/Store/FavoritesStore.swift
```diff
@@ -47,11 +48,17 @@ public class FavoritesStore: Store<UserFavorite> {
...
+            return Promise(error: FavoritesErrors.noAccount)
+        }
         let fav = UserFavorite()
         fav.productId = forProduct.id
         fav.accountId = accountId
-        self.createEntry(entry: fav, completion: completion)
+        return self.createEntry(entry: fav)
+    }
+    
+    enum FavoritesErrors : Error {
```
bhariharan: Maybe we should consolidate all errors in one class.

BaristaSwift/Consumer/Consumer/AppDelegate.swift
```diff
@@ -54,7 +55,8 @@ class AppDelegate : UIResponder, UIApplicationDelegate
     
     override init() {
         super.init()
-        
+        SalesforceSwiftLogger.setLogLevel(.debug)
```
bhariharan: This should be the default in a `DEBUG` build. Setting this here will enforce the same setting in a `RELEASE` build too.

BaristaSwift/Provider/Provider/AppDelegate.swift
```diff
@@ -52,6 +53,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     override init() {
         super.init()
         
+        SalesforceSwiftLogger.setLogLevel(.debug)
```
bhariharan: Similar comment as above. This shouldn't be required for a `DEBUG` build.

BaristaSwift/Common/Common/Base/Store.swift
wmathurin: reSync(syncName:x) will do a full sync the first time it's run and a delta sync subsequent times and it can be used for sync down and sync up.

PR #141, opened 2018-07-17
On updates only syncing up

wmathurin:
Now that we are logging timings around db queries/upserts and syncs, the bulk of the time is spent synching. 

To update server with local changes, there is no point doing a sync up followed by a sync down.
A sync up should suffice.

### DIFF
--- a/BaristaSwift/Provider/Provider/View Controller/ViewController.swift
+++ b/BaristaSwift/Provider/Provider/View Controller/ViewController.swift
@@ -226,5 +226,10 @@ extension ViewController: UITableViewDataSource {
         let config = UISwipeActionsConfiguration(actions: [completeAction])
         return config
     }
+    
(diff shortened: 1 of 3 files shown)


BaristaSwift/Provider/Provider/View Controller/ViewController.swift
```diff
@@ -226,5 +226,10 @@ extension ViewController: UITableViewDataSource {
         let config = UISwipeActionsConfiguration(actions: [completeAction])
         return config
     }
+    
+    func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
```
wmathurin: Unrelated cosmetic fix: the table view was showing the "delete" trailing swipe action (even though there is no handling code for it).

PR #142, opened 2018-07-17
Logout handling

wmathurin:
The application has a bunch of singletons (one per object type).
1) They were holding on to their SmartStore and SmartSync instances.
2) Soups and syncs were only setup at init time.

Now, we don't hold on to SmartStore or SmartSync instances and we setup soups and syncs on first access to store/sync manager following a logout.

NB: the app as it is would not support multiple users logged in

### DIFF
--- a/BaristaSwift/Common/Common/Base/Store.swift
+++ b/BaristaSwift/Common/Common/Base/Store.swift
@@ -33,6 +33,9 @@ import SmartSync
 import PromiseKit
 
 public class Store<objectType: StoreProtocol> {
+    
+    private var soupInitialized : Bool = false
+    private var syncsInitialized : Bool = false
 
     private final let pageSize: UInt = 100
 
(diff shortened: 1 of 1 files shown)


BaristaSwift/Common/Common/Base/Store.swift
```diff
@@ -33,20 +33,27 @@ import SmartSync
...
+    private final let soupName: String = objectType.objectName;
+    
     public let sqlQueryString: String = SFRestAPI.soqlQuery(withFields: objectType.createFields, sObject: objectType.objectName, whereClause: nil, groupBy: nil, having: nil, orderBy: [objectType.orderPath], limit: 100)!
     
     public let queryString: String = "SELECT \(objectType.selectFieldsString()) FROM {\(objectType.objectName)} WHERE {\(objectType.objectName):\(Record.Field.locallyDeleted.rawValue)} != 1 ORDER BY {\(objectType.objectName):\(objectType.orderPath)} ASC"
     
     public init() {
-        let soupName = objectType.objectName
-        
+        NotificationCenter.default.addObserver(self, selector: #selector(handleUserWillLogout), name: NSNotification.Name(rawValue: "SFNotificationUserWillLogout") /* kSFNotificationUserWillLogout */, object: nil)
```
wmathurin: @trooper2013 What would be a clean way to reference `kSFNotificationUserWillLogout`.
Right now I'm doing `NSNotification.Name(rawValue: "SFNotificationUserWillLogout")`

trooper2013: @wmathurin  you should be able to do this.
``` 
NotificationCenter.default.addObserver(self, selector: #selector(handleUserWillLogout), name: NSNotification.Name(kSFNotificationUserWillLogout), object: nil)
```

PR #152, opened 2018-08-25
Swift Modernization

brandonpage:


iOSIDPTemplate/Authenticator/AppDelegate.swift
```diff
@@ -37,6 +37,8 @@ class AppDelegate : UIResponder, UIApplicationDelegate
         super.init()
         _ = SalesforceSwiftSDKManager.initSDK()
         
+        SalesforceSwiftSDKManager.shared().isIdentityProvider = true;
```
brandonpage: This got lost in Raj's refactor.

PR #201, opened 2019-06-25
Modernizing SmartSyncExplorerReactNative

wmathurin:
- using react-navigation instead of navigation-experimental (nb: using react-navigation 2.x because react-navigation 3.x doesn't work with mobile sdk because of androidx)
- using FlatList instead of the deprecated ListView
- using `class xxx extends React.Component` instead of `createReactClass` for all components
-  moved styles to its own file (also cleaned them)
- moved NavImgButton class to its own file
- made some ui changes in detail screen (big save / small delete buttons as opposed to the opposite)

### DIFF
--- a/ReactNativeTemplate/app.js
+++ b/ReactNativeTemplate/app.js
@@ -95,4 +95,3 @@ export const App = createStackNavigator({
         screen: ContactListScreen
     }
 });
-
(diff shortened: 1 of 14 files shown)


wmathurin:
Here is the detail screen with big save button and small delete button.
Screenshot 2019-06-25 17 58 17

PR #203, opened 2019-07-16
Safe area inset for hybrid local template app

wmathurin:
Before:
Screenshot 2019-07-15 18 08 47

After:
Screenshot 2019-07-15 18 55 53

### DIFF
--- a/HybridLocalTemplate/index.html
+++ b/HybridLocalTemplate/index.html
@@ -2,10 +2,17 @@
 <html>
 <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no">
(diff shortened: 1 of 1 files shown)


HybridLocalTemplate/index.html
```diff
@@ -2,10 +2,17 @@
 <html>
 <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no">
+    <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover">
     <link href="ratchet.css" rel="stylesheet">
```
wmathurin: I think we should retire ratchet.css (it has not been updated in 4 years). But that's for another PR.

PR #205, opened 2019-07-22
Update Kotlin and remove deprecated ndk.

brandonpage:


brandonpage:
@bhariharan No kotlin and Wolf already removed the ndk line a month ago for those templates.

PR #206, opened 2019-07-25
Moving to React Native 0.60.4

wmathurin:


ReactNativeTemplate/android/app/build.gradle
```diff
@@ -86,26 +86,36 @@ def enableSeparateBuildPerCPUArchitecture = false
...
+ */
+def useIntlJsc = false
+
 android {
-    compileSdkVersion 28
-    buildToolsVersion '28.0.3'
+    compileSdkVersion rootProject.ext.compileSdkVersion
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
```
bhariharan: This is likely to fail on some `API 21` and `API 22` devices - they should be capable of running `Java 1.6` if I'm not mistaken.

ReactNativeTemplate/android/build.gradle
```diff
@@ -1,6 +1,14 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+  ext {
+    buildToolsVersion = "28.0.3"
+    minSdkVersion = 21
+    compileSdkVersion = 28
+    targetSdkVersion = 28
+    supportLibVersion = "28.0.0"
```
bhariharan: Shouldn't this be the same as `buildToolsVersion`?

SmartSyncExplorerReactNative/android/app/build.gradle
```diff
@@ -87,26 +87,36 @@ def enableSeparateBuildPerCPUArchitecture = false
...
+ */
+def useIntlJsc = false
+
 android {
-    compileSdkVersion 28
-    buildToolsVersion '28.0.3'
+    compileSdkVersion rootProject.ext.compileSdkVersion
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
```
bhariharan: Same comment as above. Do we need this?

ReactNativeTemplate/android/app/build.gradle
wmathurin: I ran their CLI to generate an application and copied over their build.gradle.

ReactNativeTemplate/android/build.gradle
wmathurin: I don't think there is a 28.0.3 version of the support lib (see https://developer.android.com/topic/libraries/support-library/revisions).

bhariharan:
Cool. LGTM!

PR #211, opened 2019-08-28
Basic swift template optimized for iOS 13

wmathurin:
Basic iOS swift template that uses Swift UI + Combine
Based on https://github.com/trooper2013/SalesforceMobileSDK-SampleApp-iOS13 but using Future instead of AnyPublisher.

**We have yet to decide whether we want that in 7.3 or 8.0.**
But it's good to have the PR out for review


wmathurin:
@bhariharan what do you mean when you say "added to the list the release script handles" ?

bhariharan:
The release script goes through and replaces `#dev` with the version number of the release in `package.json`, right? How does it know which templates to go to? Isn't that a list that exists somewhere?

wmathurin:
Thanks @bhariharan  - totally forgot about it.
Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/211/commits/057098bdf4765b6bc95a93c47f893a3ec7a683e3).

wmathurin:
I'll merge this now. I'll refine it further (e.g. start using SceneDelegate) in another PR (will be easier to review that way).

PR #215, opened 2019-09-19
Renaming SmartSync to MobileSync

wmathurin:


wmathurin:
Successfully ran test_force to generate and build all app types.

PR #217, opened 2019-09-21
Basic offline first template

wmathurin:
For the interesting code (SmartStore and MobileSync combine-friendly extensions) see https://github.com/forcedotcom/SalesforceMobileSDK-Templates/commit/253223443908017dbb9540e8bf86422c0f0f88b9


iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
```diff
@@ -0,0 +1,63 @@
...
+ */
+enum QueryError: Error {
+    case error(_ error: Error?)
+    case unknown
+}
+
+/**
+ SmartStore extension that adds a query:smartSql method returning a Combine Publisher
+ */
+extension SmartStore {
```
wmathurin: Extension to get a Future when running a query (also hides QuerySpec / page size and page index).

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/MobileSyncExtension.swift
```diff
@@ -0,0 +1,55 @@
...
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import MobileSync
+import Combine
+
+/**
+ SyncManager extension
+ */
+extension SyncManager {
```
wmathurin: Extension to get a Future when running a sync (hides intermediate sync state updates).

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    /**
+        Runs a query and return a Future Publisher to consume the result
+        @param smartSql:String the query smart sql
+        @return a Future<[Any], QueryError> Publisher
+     */
+    func query(_ smartSql: String) -> Future<[Any], QueryError> {
```
trooper2013: @wmathurin It would be nice to apply the same thinking for other cud operations.   Even operations like create & alter soup can benefit from a promise pattern.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
```diff
@@ -0,0 +1,63 @@
...
+        Future<[Any], QueryError> { promise in
```
trooper2013: Is it possible that we can add more type information to the return type instead of Any?

trooper2013: @wmathurin We should also see if theres a way to support unmarshalling these objects into Types using Encodables. For Rest API I will be adding such support.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/MobileSyncExtension.swift
```diff
@@ -0,0 +1,55 @@
...
+        Runs a sync and return a Future Publisher
+        @param synaName:String the name of the sync to (re)run
+        @return a Future<Bool, Error> Publisher: boolean indicates completion/failure, error is only returned if the sync could not be started (e.g. invalid name)
+     */
+    func reSync(named syncName: String) -> Future<Bool, Error> {
+        Future<Bool, Error> { promise in
+            
+            do {
+                try self.reSync(named: syncName) { (state) in
+                    if (state.isDone()) {
```
trooper2013: nit: No need for the parenthesis with if.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/MobileSyncExtension.swift
```diff
@@ -0,0 +1,55 @@
...
+                        promise(.success(true))
+                    } else if (state.hasFailed()) {
```
trooper2013: nit: No need for the parenthesis with else if.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
```diff
@@ -0,0 +1,63 @@
...
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import SmartStore
+import Combine
+
+struct Constants {
+    static let page_size:UInt = 65536
```
trooper2013: nit: Space after page_size:

trooper2013: Can we just name it pageSize? If you are looking to namespace the constant, should use an enum instead? Im ok with this approach  as well.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
wmathurin: I only wrote methods I called in the app. May the full extension would be better done in the SDK.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SmartStoreExtension.swift
wmathurin: Not sure what the signature would look like.
You could do a `select some-number-field, some-string-field, soup-element` so it would return an array of arrays where the internal arrays are of the form `[number, string, dictionary]`.
I can see the value of using an encodable for turning the dictionary into a typed object but not for the rest of the result set. So it might be better done the `.tryMap` of the code calling `query`.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/AppDelegate.swift
```diff
@@ -0,0 +1,140 @@
+/*
+ Copyright (c) 2015-present, salesforce.com, inc. All rights reserved.
```
bhariharan: Everything should be `2019-present`.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/InitialViewController.swift
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright (c) 2014-present, salesforce.com, inc. All rights reserved.
```
bhariharan: `2019-present`.

iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/main.swift
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright (c) 2015-present, salesforce.com, inc. All rights reserved.
```
bhariharan: `2019-present`.

iOS13OfflineNativeSwiftTemplate/template.js
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2016-present, salesforce.com, inc.
```
bhariharan: `2019-present`.

wmathurin:
@trooper2013 I'm going to merge that in.
We can revisit once we have swift overlays in the SDK.

PR #228, opened 2019-11-04
Store and syncs need to be configured during a user switch also

wmathurin:
Also cleaning up template.js files


iOS13OfflineNativeSwiftTemplate/iOS13OfflineNativeSwiftTemplate/SceneDelegate.swift
```diff
@@ -94,7 +90,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
    }
    
    func setupRootViewController() {
```
wmathurin: This method is called from two places:
```swift
        AuthHelper.registerBlock(forCurrentUserChangeNotifications: {
           self.resetViewState {
               self.setupRootViewController()
           }
        })
```
And
```swift
        AuthHelper.loginIfRequired {
           self.setupRootViewController()
       }
```

PR #229, opened 2019-11-05
Disable dark mode for iOS react apps

bbirman:


bhariharan:
Why are we disabling dark mode in our `ReactNative` apps?

PR #234, opened 2019-11-14
Add template for iOS encrypted push notifications

bbirman:
This is created off of iOSNativeSwiftTemplate with some changes
- Removed commented out customization options in the app delegate that weren't related to push notifications
- Add notification extension and related push entitlements


iOSNativeSwiftEncryptedNotificationTemplate/NotificationServiceExtension/NotificationService.swift
```diff
@@ -0,0 +1,55 @@
...
+    override func serviceExtensionTimeWillExpire() {
+        // Called just before the extension will be terminated by the system.
+        // Use this as an opportunity to deliver your "best attempt" at modified content, otherwise the original push payload will be used.
+        handleNotification(bestAttemptContent)
+    }
+    
+    private func handleNotification(_ notificationContent: UNMutableNotificationContent?) {
+        if let contentHandler = contentHandler, let notificationContent = bestAttemptContent {
+            do {
+                try SFSDKPushNotificationDecryption.decryptNotificationContent(notificationContent)
```
bbirman: The exciting part!

iOSNativeSwiftTemplate/iOSNativeSwiftTemplate.xcodeproj/project.pbxproj
```diff
@@ -144,7 +144,7 @@
 			attributes = {
 				LastSwiftUpdateCheck = 0710;
 				LastUpgradeCheck = 0940;
-				ORGANIZATIONNAME = Salesforce;
+				ORGANIZATIONNAME = iOSNativeSwiftTemplateOrganizationName;
```
bbirman: @wmathurin it looked like this one didn't match up to its template.js, let me know if I misunderstood!

iOSNativeSwiftEncryptedNotificationTemplate/template.js
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2016-present, salesforce.com, inc.
```
wmathurin: 2019 ;-)

iOSNativeSwiftTemplate/iOSNativeSwiftTemplate.xcodeproj/project.pbxproj
wmathurin: Good catch.

iOSNativeSwiftEncryptedNotificationTemplate/Podfile
```diff
@@ -0,0 +1,29 @@
+platform :ios, '12.2'
+
+project 'EncryptedNotificationTemplate.xcodeproj'
+target 'EncryptedNotificationTemplate' do
+    source 'https://cdn.cocoapods.org/'
+    use_frameworks!
+
+    pod 'SalesforceSDKCommon', :path => 'mobile_sdk/SalesforceMobileSDK-iOS'
```
bhariharan: You don't really need these 4 entries - `MobileSync` should transitively pull the others in through its `Podfile`.

wmathurin: We do because we want them pulled from a local path.
Otherwise, it will try to pull to one published.

bhariharan: Ah yes, I forgot that the local ones aren't updated to pull transitively locally as well.

wmathurin: Our libraries happen to live in one directory but it's typically not the case. So there would be no way for cocoapod to know what local pod of B to get when using a local pod of A (assuming A depends on B) or even to know if that's what one wants.

bhariharan: Makes sense.

PR #240, opened 2019-12-09
Limit commented out code

wmathurin:
Instead of having lines and lines of commented out codes that we often forget to refactor whenever we make changes in the SDK, the example code now lives in small private methods. Only the calls to those privates methods are commented out.


wmathurin:
The PR right now only fixes one template, I'll update the other templates if everybody agrees on this approach.

PR #242, opened 2019-12-11
Fix Android IDP Template User List.

brandonpage:


AndroidIDPTemplate/app/src/com/salesforce/samples/salesforceandroididptemplateapp/MainActivity.kt
```diff
@@ -115,7 +115,7 @@ class MainActivity : SalesforceActivity() {
 
         // Displays list of users available.
         usersListView?.adapter = ArrayAdapter(this,
-                android.R.layout.simple_selectable_list_item, arrayListOf(buildListOfUsers()))
+                android.R.layout.simple_selectable_list_item, buildListOfUsers())
```
brandonpage: `buildListOfUsers()` already returns a list.  Prior to this change all users would be in a list in a single cell with a string like "[user 1], [user 2], [add new user]".  Now each is displayed in its own list item.

PR #258, opened 2020-03-31
Move MobileSyncExplorerSwift to Combine/SwiftUI

bbirman:
image


MobileSyncExplorerSwift/MobileSyncExplorerSwift/ViewControllers/RootViewController.swift
```diff
@@ -1,315 +0,0 @@
...
-    
-    required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
-    }
-    
-    override func viewDidLoad() {
-        super.viewDidLoad()
-        self.view.backgroundColor = UIColor.salesforceSystemBackground
-        
-        NotificationCenter.default.addObserver(self,
```
bbirman: I tested the new code with an alert up and passcode went over it so I didn’t bring this over, let me know if there’s anything else I should verify

MobileSyncExplorerSwift/MobileSyncExplorerSwift/Views/ContactList.swift
```diff
@@ -0,0 +1,238 @@
...
+import MobileSync
+import Combine
+
+struct ContactListView: View {
+    @ObservedObject private var sObjectDataManager: SObjectDataManager = SObjectDataManager(dataSpec: ContactSObjectData.dataSpec()!)
+    @State private var searchTerm: String = ""
+    @State private var dim = false
+
+    init() {
+        self.sObjectDataManager.syncUpDown()
```
bbirman: Putting this in onAppear doesn’t work because then the network call is triggered every time going from detail back to list, let me know if there’s a better place for it

MobileSyncExplorerSwift/MobileSyncExplorerSwift/Views/ContactList.swift
```diff
@@ -0,0 +1,238 @@
...
+        searchBar.delegate = context.coordinator
+        return searchBar
+    }
+
+    func updateUIView(_ uiView: UISearchBar, context: UIViewRepresentableContext<SearchBar>) {
+        uiView.text = text
+    }
+}
+
+struct InspectorViewControllerWrapper: UIViewControllerRepresentable {
```
bbirman: Since these are generic wrappers to use the view controllers in SwiftUI, I’m wondering if we should add them to the SDK for convenience

bbirman:
@wmathurin, cool, will do!

PR #279, opened 2020-07-11
Add notification list to MobileSyncExplorerSwift

bbirman:
image


MobileSyncExplorerSwift/MobileSyncExplorerSwift/Views/NotificationList.swift
```diff
@@ -0,0 +1,97 @@
...
+    @ObservedObject var model: NotificationListModel
+    var sObjectDataManager: SObjectDataManager
+
+    var body: some View {
+        List {
+            ForEach(model.notifications, id: \.id) { notification in
+                VStack {
+                    if notification.targetId.starts(with: "003") { // Only contacts are tappable
+                        NavigationLink(destination: ContactDetailView(contactId: notification.targetId, sObjectDataManager: self.sObjectDataManager, onAppear: {
+                            self.model.markNotificationRead(notificationId: notification.id)
```
bbirman, line 59: Passing this as a work around because .simultaneousGesture doesn't work with navigation links in lists

MobileSyncExplorerSwift/MobileSyncExplorerSwift/NotificationListModel.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    func unreadCount() -> Int {
+        return notifications.filter { !$0.read }.count
+    }
+
+    func fetchNotifications() {
+        let builder = FetchNotificationsRequestBuilder.init()
+        builder.setBefore(Date())
+        builder.setSize(5)
+        let request = builder.buildFetchNotificationsRequest(SFRestDefaultAPIVersion)
```
wmathurin, line 119: Is there any setup needed on the server to produce those notifications?

bbirman: Yes, creating a custom notification type and then tying it to the connected app. I'd been thinking about something we could push to the org but it'd still be dependent on the connected app. Sue and I were also talking about outlining the other steps in the post she's working on

wmathurin: 👍

PR #307, opened 2021-02-17
Templates using wrong @types/react version in dependencies

wmathurin:


bhariharan (approved):

wmathurin:
Getting errors like [that](https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes) when using 16.x instead of 17.x, closing while investigating further.

PR #337, opened 2022-02-07
Add recent contacts widget for MobileSyncExplorerSwift

bbirman:
https://user-images.githubusercontent.com/840431/152848237-de1fdab9-6a17-47f1-ae65-5542fccd9764.mov

### DIFF
--- a/MobileSyncExplorerSwift/RecentContacts/RecentContactsWidget.swift
+++ b/MobileSyncExplorerSwift/RecentContacts/RecentContactsWidget.swift
@@ -18,7 +18,7 @@ struct Provider: TimelineProvider {
     }
     
     func currentEntry(context: Context) -> ContactsEntry {
-        guard let contacts = RecentContacts.shared.persistedContacts() else {
+        guard let contacts = RecentContacts.persistedContacts() else {
             return placeholder(in: context)
         }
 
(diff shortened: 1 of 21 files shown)


MobileSyncExplorerSwift/RecentContacts/RecentContactsWidget.swift
```diff
@@ -0,0 +1,141 @@
...
+//  Created by Brianna Birman on 1/21/22.
+//  Copyright © 2022 MobileSyncExplorerSwiftOrganizationName. All rights reserved.
+//
+
+import WidgetKit
+import SwiftUI
+
+struct Provider: TimelineProvider {
+    func placeholder(in context: Context) -> ContactsEntry {
+        let contacts = [ContactSummary(id: "1", firstName: "Michelle", lastName: "Kim"),
```
wmathurin, line 14: What are the hardcoded values for?

bbirman: The placeholder part will show a redacted version of this view as a loading stencil so it won't actually show the data, and then snapshot is also using it. Snapshot is used for the discovery view (screenshot below), so this will show even if the user isn't logged in, but it'd also be nice to show real data when they are logged in, could add a check for that
Screen Shot 2022-02-07 at 10 54 25 AM

wmathurin (approved):

PR #346, opened 2022-09-19
Fixing set_version.sh script to include MobileSyncExplorerKotlinTemplate

wmathurin:


gkotula (approved):
Thanks for the quick fix on this

PR #350, opened 2022-12-13
Add Kotlin Dependency and update Gradle

brandonpage:
TODO:
- [x] Update Kotlin for hybrid apps - Done in [CordovaPlugin PR](https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/pull/611)
- [x] Fix MobileSyncExplorerReactNative compilation (this may just be my environment) - JVM needed more memory.

### DIFF
--- a/AndroidNativeTemplate/build.gradle
+++ b/AndroidNativeTemplate/build.gradle
@@ -5,7 +5,8 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:7.2.1'
+    classpath 'com.android.tools.build:gradle:7.3.1'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0"
   }
 }
 
--- a/AndroidIDPTemplate/app/build.gradle
+++ b/AndroidIDPTemplate/app/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'kotlin-android'
 
 dependencies {
   implementation project(':libs:MobileSync')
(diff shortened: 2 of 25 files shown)


AndroidNativeTemplate/build.gradle
```diff
@@ -5,7 +5,8 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:7.2.1'
+    classpath 'com.android.tools.build:gradle:7.3.1'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.6.0"
```
brandonpage, line 9: Since Kotlin isn't used in this project I figured I would add it as the min version needed, just to ensure that remains true.  Gradle in our main repo _should_ throw errors if we use new features, but I figured this couldn't hurt as an extra safety net.

bhariharan (approved):

wmathurin (approved):
LGTM
What needs to be done for the generated hybrid apps?

brandonpage:
> LGTM What needs to be done for the generated hybrid apps?

That is a great question.  `cdv-gradle-config.json` specifies Kotlin 1.5.21 but I am not sure where it is coming from.  It doesn't seem like they should compile, but they do.  Comcast is doing work in my neighborhood so I mostly just wanted to get this PR out before the internet goes out.  I will update the PR when I get that fixed.

wmathurin:
The plugin can override values that come from the cordova json files - we do that already [here](https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/blob/dev/plugin.xml#L161), it's just a matter of figuring out the property name.

wmathurin (approved):

PR #361, opened 2023-06-15
Bump iOS deployment target to 15.0

bbirman:


MobileSyncExplorerReactNative/ios/Podfile
```diff
@@ -3,7 +3,7 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
 require_relative '../mobile_sdk/SalesforceMobileSDK-iOS/mobilesdk_pods'
 
 
-platform :ios, '14.0'
+platform :ios, '15.0'
 
 project 'MobileSyncExplorerReactNative.xcodeproj'
```
bbirman, line 8: @wmathurin I wasn't sure why [this part](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/6f55cec69dc07d44c51dca45998be10448b4954e/MobileSyncExplorerReactNative/ios/Podfile#L52) of the file has 12.0, should I change it to 15 too or leave it?

wmathurin: That was a fix to make it work with Xcode 14.3 (suggested [here](https://github.com/facebook/react-native/pull/36759#issuecomment-1493710771)). We won't need it once we upgrade to React Native 71.

wmathurin (approved):

MobileSyncExplorerReactNative/ios/Podfile
wmathurin: Also would you mind fixing that [typo](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/MobileSyncExplorerReactNative/ios/Podfile#L49) since you are in those files :-)

brandonpage (approved):

MobileSyncExplorerReactNative/ios/Podfile
wmathurin: Oops I copied/pasted the same typo in the other react native templates (e.g. [here](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/ReactNativeTemplate/ios/Podfile#L49)).

PR #360, opened 2023-06-12
Moving package name to build.gradle

wmathurin:
Addressing deprecation message during builds:

> package="com.salesforce.androidsdk" found in source AndroidManifest.xml: /home/circleci/SalesforceMobileSDK-Package/tmp20230608T035347.445/androidnativekotlin/mobile_sdk/SalesforceMobileSDK-Android/libs/SalesforceSDK/AndroidManifest.xml.
Setting the namespace via a source AndroidManifest.xml's package attribute is deprecated.
Please instead set the namespace (or testNamespace) in the module's build.gradle file, as described here: https://developer.android.com/studio/build/configure-app-module#set-namespace

PR #363, opened 2023-06-16
Fix MobleSyncExplorerKotlinTemplate compilation.

brandonpage:


wmathurin:
[template.js](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/MobileSyncExplorerKotlinTemplate/template.js#L59) (which replaces template package name by user provider package name) expects `com.salesforce.mobilesyncexplorerkotlintemplate`. 

The problem is that my recent [PR](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/360) stopped doing the replacements in the AndroidManifest.xml.

I'll send a PR to fix the template.js scripts and close this PR.

wmathurin:
@brandonpage Here is the [fix](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/364) (I did not have to change template.js, I simply used relative names for the application and activity in AndroidManifest.xml).

PR #364, opened 2023-06-19
Not having full package name in AndroidManifest.xml

wmathurin:
In the past template.js was replacing the template package name by the user provided package name in AndroidManifest.xml, so having the full package name in there was not a problem.
With https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/360, I moved the package name to build.gradle and modified template.js to no touch AndroidManifest.xml.


JohnsonEricAtSalesforce (approved):
⭐️

PR #367, opened 2023-06-22
Replacing app name in all places for MobileSyncExplorerKotlinTemplate

wmathurin:


MobileSyncExplorerKotlinTemplate/app/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,5 @@
 <resources>
-    <string name="app_name">Mobile Sync Explorer Kotlin Template</string>
+    <string name="app_name">MobileSyncExplorerKotlinTemplate</string>
```
wmathurin, line 2: The name without spaces is used in other places (e.g. settings.gradle), so just using non-spaced name everywhere and searching for it in template.js.

brandonpage (approved):

PR #369, opened 2023-09-15
@W-14125965: [MSDK Android] Cannot Compile `MobileSyncExplorerKotlinTemplate` After MSDK Gradle 8 And Kotlin DSL Updates

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

This resolves an error when compiling the MSDK Android template apps, such as `MobileSyncExplorerKotlinTemplate`, where the publishing module Gradle Script Plugin introduced with the MSDK Gradle 8 and Kotlin DSL update cannot be accessed.

The error originates in how the template-based app includes MSDK in its Gradle workspace.  The template-based app uses a separate Gradle include for specific MSDK modules.  This doesn't include the overall MSDK workspace and leaves out anything in MSDK's build source.

As it turns out, simply migrating to a full composite build with MSDK greatly simplifies the configuration.  MSDK builds with its full Gradle workspace, including the custom Gradle script plugin for the publishing module.  This will likely make the Gradle integration between the app and MSDK much more robust in the future.

Also, since the app and MSDK are building together in Gradle from source, they will need to maintain the same Gradle and Android plugin versions.  So, this gets those in sync as well.

There's more.  Some React Native community libraries are not yet compatible with Android Gradle Plugin 8 and will not allow the workspace to sync.  So, the Gradle scripts now include a check for the missing `namespace` property and apply a placeholder where needed, which is a work-around in the community.  Note that the next version of React Native itself will migrate to Gradle 8 and Android Gradle Plugin 8, so the authors have posted a requirement that all community contributions will have to be compatible at that time.

I created a follow-up work item to migrate all these Gradle scripts to Kotlin DSL, since that would have made this set of changes difficult to follow in Github.  See [W-14149830](https://gus.lightning.force.com/a07EE00001aoVYYYA2)


AndroidIDPTemplate/app/build.gradle
```diff
@@ -2,13 +2,13 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 dependencies {
-  implementation project(':libs:MobileSync')
+    implementation 'com.salesforce.mobilesdk:MobileSync:11.0.0'
```
wmathurin, line 5: If we do it this way it won't dev version of the template won't use the dev version of the Mobile SDK and we won't be able to detect regressions until we tag the release.

wmathurin: Is the idea to do:
- get lib from maven central in production 
- get lib through a composite build in dev
In that case we should tweak our setversion.sh / install.js / settings.gradle files between prod and dev.
In prod, we should omit the `includeBuild(new File(settingsDir, 'mobile_sdk/SalesforceMobileSDK-Android' from settings.gradle and not clone the SalesforceMobileSDK-Android repo in install.js.

JohnsonEricAtSalesforce:
It's worth noting I opened every template in Android Studio to verify Gradle sync and a run of the app on device will function correctly.  It does take a little while!

AndroidIDPTemplate/app/build.gradle
```diff
@@ -2,13 +2,13 @@ apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 
 dependencies {
-  implementation project(':libs:MobileSync')
+    implementation 'com.salesforce.mobilesdk:MobileSync'
```
JohnsonEricAtSalesforce, line 5: @wmathurin - I opted to remove the version specifier so perhaps it is clearer this will be resolved from the local contents of the Gradle composite build via `includeBuild`.

MobileSyncExplorerReactNative/android/build.gradle
```diff
@@ -33,3 +33,25 @@ allprojects {
...
+ *
+ * Until that time, this work-around applies a temporary `namespace` only as
+ * needed.
+ *
+ * https://github.com/react-native-community/discussions-and-proposals/issues/671
+ */
+subprojects {
+  afterEvaluate {
+    if (android.namespace == null) {
+      logger.warn("Applying a Gradle 8 and Android Gradle Plugin 8 compatible namespace to Android project named '" + project.name + "'.  Consider upgrading to a new version, if possible.")
```
JohnsonEricAtSalesforce, line 53: This is the magic that lets React Native libraries missing Android Gradle Plugin's required `namespace` attribute to build.  Beyond the community post I read, I added a check so it's applied only when needed and a log to hint where it was applied.

brandonpage (approved):

PR #372, opened 2023-09-28
@W-14149830: [MSDK Android] Convert Template Apps To Gradle Kotlin DSL

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

This prepares `SalesforceMobileSDK-Templates` for most of its Gradle Groovy DSL to be converted to Kotlin DSL.  See the [companion pull request](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2464).

Unfortunately, it wasn't possible to coerce React Native nor its community libraries to integrate with Gradle 8 and the Android Gradle Plugin 8, so this rolls that back to 7 here.  The two repositories must remain in sync since they are a composite build.

React Native will be moving to Gradle 8 in the near future, though the timing of community library updates is less clear.

In the end, all the Groovy DSL that is not derived from React Native templates can be migrated to Kotlin DSL, so a large body of work is still possible.  Yay!

All of the templates apps will sync, compile and run based on testing.  Thanks for reading and let me know what else might remain.


AndroidIDPTemplate/app/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+
+    buildTypes {
+      debug {
+         enableAndroidTestCoverage = true
+      }
+   }
+
+  sourceSets {
+
+        getByName("main") {
```
wmathurin, line 29: NIT indentation

AndroidNativeKotlinTemplate/gradle.properties
```diff
@@ -1,6 +1,6 @@
 cdvCompileSdkVersion=android-33
 cdvMinSdkVersion=24
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
```
wmathurin, line 4: Why the change?

AndroidIDPTemplate/app/build.gradle.kts
JohnsonEricAtSalesforce: I just checked that his is Android Studio's default.  Did you mean the extra newline above this?

AndroidNativeKotlinTemplate/gradle.properties
JohnsonEricAtSalesforce: The build analyzer suggested it with a confirmation that there are no libraries in the dependency graph which require it.  I've used this quite a bit so I have some trust in the tool, but wouldn't mind reverting it, either.  The build time gain is probably negligible.

wmathurin: 👍

AndroidIDPTemplate/app/build.gradle.kts
wmathurin: Never mind, I thought it looked off in the diff, but it isn't.

wmathurin (approved):

PR #373, opened 2023-09-29
Fixing template.js scripts

wmathurin:
Should fix generating apps with forcedroid.


JohnsonEricAtSalesforce (approved):
⭐️ Thanks, @wmathurin!

PR #374, opened 2023-10-04
New iOS template that uses SPM

wmathurin:


iOSNativeSwiftPackageManagerTemplate/install.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
```
wmathurin, line 1: The only interesting file in this PR (the rest is a copy of the iOSNativeSwiftTemplate).
We read from package.json what SPM repo / branch we should be pointing to and modify the xcode project file accordingly.

wmathurin:
NB: I'll send another PR to point package.json of the new template to forcedotcom/SalesforceMobileSDK-iOS-SPM when it's finally public.

bbirman (approved):

PR #377, opened 2023-10-10
@W-14246983: [Android][Hybrid] Hybrid Templates Use Maven Central MSDK Dependencies By Default

JohnsonEricAtSalesforce:
🥁 _Ready For Review_ 🎸

  This is a companion to https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/pull/626

  This updates the hybrid templates to clone a copy of MSDK-Android for use in their composite build when the version is not a tagged release.

### DIFF
--- a/HybridLocalTemplate/install.js
+++ b/HybridLocalTemplate/install.js
@@ -14,9 +14,9 @@ for (var sdkDependency in packageJson.sdkDependencies) {
     // Exclude tagged SalesforceMobileSDK-Android releases so it's only a source dependency pre-release.
     if (sdkDependency == 'SalesforceMobileSDK-Android' && branch.match(/v\d+\.\d+\.\d+/)) {
        if (fs.existsSync(targetDir)) {
-           console.log('SalesforceMobileSDK-Android is release version.  Warning: Its published artifacts will not be used since sources are already in ' + targetDir + ' and will be used.  If desired, remove this directory to return to the release artifacts.');
+           console.log('SalesforceMobileSDK-Android is a release version.  Warning: Its published artifacts will not be used since sources are already in ' + targetDir + ' and will be used.  If desired, remove this directory to return to the release artifacts.');
        } else {
-           console.log('SalesforceMobileSDK-Android is release version.  Skipping its source dependency and using published artifacts since it\'s only a source dependency for pre-release versions.');
+           console.log('SalesforceMobileSDK-Android is a release version.  Skipping its source dependency and using published artifacts since it\'s only a source dependency for pre-release versions.');
        }
        continue;
     }
--- a/HybridLocalTemplate/install.js
+++ b/HybridLocalTemplate/install.js
@@ -16,7 +16,7 @@ for (var sdkDependency in packageJson.sdkDependencies) {
        if (fs.existsSync(targetDir)) {
            console.log('SalesforceMobileSDK-Android is release version.  Warning: It\'s published artifacts will not be used since sources are already in ' + targetDir + ' and will be used.  If desired, remove this directory to return to the release artifacts.');
        } else {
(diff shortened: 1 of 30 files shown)


HybridLocalTemplate/install.js
```diff
@@ -10,6 +10,13 @@ for (var sdkDependency in packageJson.sdkDependencies) {
     var repoUrlWithBranch = packageJson.sdkDependencies[sdkDependency];
     var parts = repoUrlWithBranch.split('#'), repoUrl = parts[0], branch = parts.length > 1 ? parts[1] : 'master';
     var targetDir = path.join('mobile_sdk', sdkDependency);
+
+    if (sdkDependency == 'SalesforceMobileSDK-Android' && branch.match(/v\d+\.\d+\.\d+/)) {
+        console.log('Skipping dependency installation of production SalesforceMobileSDK-Android release.  It is only used for source builds in pre-production.  Production builds default to release repository artifacts.');
+        continue;
+    }
```
wmathurin, line 17: They might have a left over mobile sdk, maybe we should warn them like we do on line 20, or maybe move the line 20 check earlier.

wmathurin: We should do the same in the other hybrid templates: [hybrid remote](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/HybridRemoteTemplate/install.js) and [hybrid lwc](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/HybridLwcTemplate/install.js).

Where you planning to deal with native templates in this PR or another PR?

JohnsonEricAtSalesforce: Before this wraps up, each of the related pull requests will need some finalization, including updating the remaining hybrid templates, at least, and switching to formal branches.

HybridLocalTemplate/install.js
```diff
@@ -10,6 +10,17 @@ for (var sdkDependency in packageJson.sdkDependencies) {
     var repoUrlWithBranch = packageJson.sdkDependencies[sdkDependency];
     var parts = repoUrlWithBranch.split('#'), repoUrl = parts[0], branch = parts.length > 1 ? parts[1] : 'master';
     var targetDir = path.join('mobile_sdk', sdkDependency);
+
+    // Exclude tagged SalesforceMobileSDK-Android releases so it's only a source dependency pre-release.
+    if (sdkDependency == 'SalesforceMobileSDK-Android' && branch.match(/v\d+\.\d+\.\d+/)) {
+       if (fs.existsSync(targetDir)) {
```
JohnsonEricAtSalesforce, line 16: @wmathurin - Take a look at this for https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/377#discussion_r1353763217

wmathurin: Looks good. 

In terms of messaging, what do you think about doing the following:
```javascript
console.log(`Using ${skdDependency} at ${branch}`);
if (sdkDependency == 'SalesforceMobileSDK-Android' && branch.match(/v\d+\.\d+\.\d+/)) {
    if (fs.existsSync(targetDir)) {
        console.log(`Warning: Published artifacts will not be used since sources are already in ${targetDir}. If desired, remove this directory to return to the release artifacts.`);    
    }
   continue; // Getting SDK from Maven Central
}
```

HybridLocalTemplate/install.js
wmathurin: 👍

HybridLocalTemplate/install.js
wmathurin: @JohnsonEricAtSalesforce Maybe :arrow_up: in the install.js ?

wmathurin (approved):

PR #395, opened 2024-03-13
@W-14584260: [iOS] OTPs Sample App

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🤘🏻

  This updates the sample app with the ability to navigate to other login, registration and forgot password flows.  The password-less login view one-time-passcode is the only new flow that will be implemented in this update.

  There's lots of new SwiftUI concepts at play for the first time here.

   @brandonpage - There's one problem with navigation bars in UIKit Vs. SwiftUI we have to look at before we merge this one.  Look for your call to action in the code walkthrough and see if you have any ideas.


iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
 struct NativeLoginView: View {
     @Environment(\.colorScheme) var colorScheme
     
+    /// The reCAPTCHA client used to obtain reCAPTCHA tokens when needed for Salesforce Headless Identity API requests.
+    @EnvironmentObject var reCaptchaClientObservable: ReCaptchaClientObservable
+    
+    /// The navigation path.
+    @EnvironmentObject var navigationPathObservable: NavigationPathObservable
```
JohnsonEricAtSalesforce, line 38: SwiftUI Noob At Work 🚧 This pull request introduces multi-level navigation via `NavigationStack` to the template app for the first time.  So, I followed some Apple docs and created a path of custom `NavigationDestination` enums to drive the flow in and out of the navigation stack with associated values.  It sure looks cool to my rookie eyes, but I'd be game to see how the team likes it.

I was curious to maybe _not_ use an observable for this one and see how the binding to the `NavigationStack` could be simplified.  This works really well though as-is.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
...
-                            .tint(colorScheme == .dark ? .white : .blue)
-                            .opacity(0.5)
-                    }
-                    .frame(width: 30, height: 30, alignment: .topLeading)
-                    .padding(.leading)
-                    
-                } else {
-                    Spacer().frame(height: 30)
-                }
+        NavigationStack(path: $navigationPathObservable.navigationPath) {
```
JohnsonEricAtSalesforce, line 46: Here's the new `NavigationStack`.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
...
-                SalesforceManager.shared.nativeLoginManager().fallbackToWebAuthentication()
+                        .buttonStyle(.bordered)
+                        .tint(.blue)
+                        .zIndex(2.0)
+                    }.padding(.bottom, 0)
+                }
+                
+                // Other login options.
+                Button("Need to register, reset your password or login without a password?") {
+                    navigationPathObservable.navigationPath.append(.NativeRequestOtpView)
```
JohnsonEricAtSalesforce, line 160: Github is showing a lot more difference here than there really is.  That might be whitespace related with the new `NavigationStack` element.  However, the only change is really the addition of the "Other login options" button here.  That's the app's way of letting the user escape from username and password login to register, forgot password and password-less one-time-passcode login.  Only the latter of those is currently implemented.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
...
+                Button("Looking for Salesforce Log In?") {
+                    SalesforceManager.shared.nativeLoginManager().fallbackToWebAuthentication()
+                }
             }
-        }.background(Gradient(colors: [.blue, .cyan, .green]).opacity(0.6))
+            .background(
+                Gradient(colors: [.blue, .cyan, .green]).opacity(0.6)
+            )
             .blur(radius: self.isAuthenticating ? 2.0 : 0.0)
+            .navigationDestination(for: NavigationDestination.self) { navigationDestination in
```
JohnsonEricAtSalesforce, line 174: Here's the new `navigationDestination` modifier for the `NavigationStack`.  This declarative navigation DSL is actually a lot of fun (Though Xcode's SwiftUI "Canvas" crashes _a lot_.)

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
...
+                switch (navigationDestination) {
+                    
+                case .NativeRequestOtpView:
+                    NativeRequestOtpView()
+                    
+                case .NativeSubmitOtpView (
+                    let otpIdentifier,
```
JohnsonEricAtSalesforce, line 181: It's further down in the diff, but notice the associated values on the `NavigationDestination` Swift enum let the OTP verification parameters get passed through super easy and type safe.  I thought this really decoupled the views and feels like the pattern Android Navigation uses when declaring navigation parameters (though that's still in XML - ew.)

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,134 +31,183 @@ import SalesforceSDKCore
...
 }
 
 #Preview {
     NativeLoginView()
 }
+
+///
+/// The available navigation destinations.
+///
+enum NavigationDestination: Hashable {
```
JohnsonEricAtSalesforce, line 200: Here's the `NavigationDestination` enum that is the core navigation model.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginViewFactory.swift
```diff
@@ -28,11 +28,35 @@
...
-        let view = UIHostingController(rootView: NativeLoginView())
-
+class NativeLoginViewFactory: NSObject {
+    
+    static func create(
+        navigationPathObservable: NavigationPathObservable,
+        reCaptchaClientObservable: ReCaptchaClientObservable
+    ) -> UIViewController {
+        let view = NativeLoginTemplateHostingController(rootView: NativeLoginView()
+            .environmentObject(navigationPathObservable)
```
JohnsonEricAtSalesforce, line 38: I thought there might be a way to add environment objects to the app or scene in such a way `NativeLoginView` would automatically pick them up, but it wouldn't bind.  So, there're added explicitly here.  As the root/content view, these objects are then available all the way down the view hierarchy.

I'd be curious to review the app and scene configuration to see if this can be improved.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginViewFactory.swift
```diff
@@ -28,11 +28,35 @@
...
+        guard let navigationController = (parent as? UINavigationController) else {
+            return
+        }
+        
+        //
+        // The Salesforce Mobile SDK provided navigation controller displays a
+        // navigation bar that is redundant with the SwiftUI navigation in the
+        // template app, so hide it.
+        //
+        // TODO: See if this should be resolved in SFMSDK. ECJ20240516
```
JohnsonEricAtSalesforce, line 58: @brandonpage - The comment covers the issue.  Paraphrased, MSDK provides a navigation bar and SwiftUI's `NavigationStack` wants to put another _under it_.  I see lots of SwiftUI devs doing work-arounds like this where their UI is getting embedded in legacy `UINavigationController` instances.  Any ideas?  This keeps the MSDK navigation bar hidden and the template looks alright, but it's certainly not ideal.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeRequestOtpView.swift
```diff
@@ -0,0 +1,202 @@
...
+            Gradient(colors: [.blue, .cyan, .green]).opacity(0.6)
+        )
+        .blur(radius: isAuthenticating ? 2.0 : 0.0)
+    }
+    
+    ///
+    /// Submits the OTP delivery request when the request button is tapped.  This submits a request to the
+    /// `init/passwordless/login`endpoint and opens the submit OTP verification view on success.
+    ///
+    private func onRequestOtpTapped() {
```
JohnsonEricAtSalesforce, line 141: Everything in this view is straight-forward SwiftUI and this is the key logic using MSDK native login manger.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeRequestOtpView.swift
```diff
@@ -0,0 +1,202 @@
...
+                    errorMessage = "Invalid username."
+                    break
+                    
+                case .unknownError:
+                    errorMessage = "An error occurred."
+                    break
+                    
+                case .success:
+                    guard let otpIdentifier = result.otpIdentifier else { return }
+                    navigationPathObservable.navigationPath.append(
```
JohnsonEricAtSalesforce, line 190: I really dig this new navigation pattern (can you tell?)

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeSubmitOtpView.swift
```diff
@@ -0,0 +1,162 @@
...
+        // Reset the error message if needed.
+        errorMessage = ""
+        
+        // Show the progress indicator.
+        isAuthenticating = true
+        
+        // Submit the request and act on the response.
+        Task {
+            // Submit the request.
+            let _ = await SalesforceManager.shared.nativeLoginManager()
```
JohnsonEricAtSalesforce, line 146: Also, really basic SwiftUI here.  the user is authenticated and returned to the app's view by MSDK at this point.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
```diff
@@ -37,10 +44,10 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         self.window = UIWindow(frame: windowScene.coordinateSpace.bounds)
         self.window?.windowScene = windowScene
         
-        ///
-        /// Fill in the values below from the connected app that was created for Native Login and
-        /// the url of your Experience Cloud community.
-        ///
```
JohnsonEricAtSalesforce, line 43: This didn't look like it needed the doc-comment.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
```diff
@@ -129,3 +163,28 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
        postResetBlock()
    }
 }
+
+@MainActor class ReCaptchaClientObservable: ObservableObject {
```
JohnsonEricAtSalesforce, line 167: When Combine first came out in 2019, I remember creating lots of these higher-level objects myself.  It's so cool to see how it's grown up in SwiftUI's context 🎉

iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
```diff
@@ -129,3 +163,28 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+            reCaptchaClient = try await Recaptcha.getClient(withSiteKey: reCaptchaSiteKey)
+        } catch let error {
+            print("Cannot get reCAPTCHA client due to an error with description '\(error.localizedDescription).'.")
+        }
+    }
+}
+
+class NavigationPathObservable: ObservableObject {
+    
+    @Published var navigationPath: [NavigationDestination] = []
```
JohnsonEricAtSalesforce, line 189: In the context of a full app, I'd love to see how this turned into popping in and back out of deep links.

JohnsonEricAtSalesforce (commented):
Here's some code commentary to get the discussion started.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeRequestOtpView.swift
```diff
@@ -0,0 +1,202 @@
...
+    @State private var otpVerificationMethod = OtpVerificationMethod.sms
+    
+    /// The user's entered username.
+    @State private var username = ""
+    
+    var body: some View {
+        VStack {
+            Spacer()
+            
+            HStack { Spacer() }
```
bbirman, line 67: Why does this Spacer need an HStack?

JohnsonEricAtSalesforce: I picked that up from `NativeLoginView`.  It seems like that was there to make the view fill the horizontal width of its parent.  It sure looked like it was to spacers: One for the vertical and one for the horizontal (Like the Twilight Zone intro).

bbirman: you might be able to cut it and the two spacers before/after the ZStack if you set something like this on the VStack
`.frame(maxWidth: .infinity, maxHeight: .infinity)`

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeSubmitOtpView.swift
bbirman: if this is the recommended way to call it, could `@discardableResult` be added to the MSDK method?

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
brandonpage: The `NavigationStack` seems a bit overkill to me when you could show either username/password or OTP flow with a simple if else instead.  It's not inherently bad, but it makes the template a bit more complex as an example.  Also, adding the nav stack appears to have broken the existing back button functionality for multiuser loginf.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
brandonpage, line 160: NIT: I think this button either needs padding or to be broken up into multiple (icon?) buttons or something.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginViewFactory.swift
brandonpage: This sounds similar to an issue @bbirman and I were discussing about a month ago.  I don't know if a bug ever got filed for it or if Brianna make any progress into the root cause.

brandonpage (changes requested):
If we are going to keep the `NavigationStack` model we at least need to fix the existing back behavior.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeRequestOtpView.swift
JohnsonEricAtSalesforce: I tried that, then realized it's that the content will "size-to-fit" centered as the rest of the layout I borrowed from native login view stands.

Screenshot 2024-03-19 at 17 06 46

The spacers are instructing the view to fill its parent edge-to-edge so it doesn't look like the above.  The native login view on dev currently wants to look more like the below.

Screenshot 2024-03-19 at 17 09 18

JohnsonEricAtSalesforce: That said, I see in the docs the modifier you sent _should_ have worked.  I'm looking again 🧐

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
JohnsonEricAtSalesforce: Agreed.  For now, I just changed it to "Use Passcode Instead".

JohnsonEricAtSalesforce:
> If we are going to keep the `NavigationStack` model we at least need to fix the existing back behavior.

After we chatted, I actually did demo a version that uses a "Cancel" button throughout the navigation instead of the custom "back arrow".  I thought it actually looked really nice so I made a video to share. 

https://github.com/forcedotcom/SalesforceMobileSDK-Templates/assets/94090596/69b31599-d96d-430c-a165-4bc95129576c

This turned into a fun update.  I like the navigation stack since it demonstrates to our app developers how the view controller passed to native login manager actually has a great deal of control over all the headless identity flows.  It can display other views and heavily tailor the user experience.  Plus, with the "Cancel" button as a persistent element the app developer knows their user can cancel authentication at any time (it's not actually a "Back" button since this view controller is presented modally).

Thoughts? I agree we don't want the template to be overly complicated yet since it will eventually have login, password-less login, forgot password and registration it should be a scalable starting point.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
brandonpage: [Passcode already has a specific meaning unrelated to this.](https://developer.salesforce.com/docs/atlas.en-us.mobile_sdk.meta/mobile_sdk/android_passcodes.htm)

brandonpage:
> > If we are going to keep the `NavigationStack` model we at least need to fix the existing back behavior.
> 
> After we chatted, I actually did demo a version that uses a "Cancel" button throughout the navigation instead of the custom "back arrow". I thought it actually looked really nice so I made a video to share.
> 
>  Simulator.Screen.Recording.-.iPhone.15.Pro.-.2024-03-19.at.19.35.26.mp4 
> This turned into a fun update. I like the navigation stack since it demonstrates to our app developers how the view controller passed to native login manager actually has a great deal of control over all the headless identity flows. It can display other views and heavily tailor the user experience. Plus, with the "Cancel" button as a persistent element the app developer knows their user can cancel authentication at any time (it's not actually a "Back" button since this view controller is presented modally).
> 
> Thoughts? I agree we don't want the template to be overly complicated yet since it will eventually have login, password-less login, forgot password and registration it should be a scalable starting point.

This looks great.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
JohnsonEricAtSalesforce: When I thought about how I'd have written this in Kotlin Compose, using conditional layout logic on a bound variable worked perfectly and was a lot simpler - just like you said.  Thanks for the nudge.

JohnsonEricAtSalesforce: That also fixed the back navigation, which I restored to the way it is on `dev` exactly.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeRequestOtpView.swift
JohnsonEricAtSalesforce: @bbirman - This all got set back to the state on `dev` in this morning's commits.

JohnsonEricAtSalesforce:
@brandonpage - I did another version and really did like a single view conditional layout better in the end.  This version only has `NativeLoginView` and it switches on an enum to determine the correct layout for the chosen identity flow.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -31,11 +31,33 @@ import SalesforceSDKCore
 struct NativeLoginView: View {
     @Environment(\.colorScheme) var colorScheme
     
-    @State private var username = ""
-    @State private var password = ""
+    /// The reCAPTCHA client used to obtain reCAPTCHA tokens when needed for Salesforce Headless Identity API requests
+    @EnvironmentObject var reCaptchaClientObservable: ReCaptchaClientObservable
+    
+    /// The layout type for the user's active identity flow, such as registration, forgot password or login
+    @State private var identityFlowLayoutType = IdentityFlowLayoutType.LoginViaUsernamePassword
```
JohnsonEricAtSalesforce, line 38: @brandonpage Without the `NavigationStack` I tried before, this enum switches between identity flow layouts here in a single view.  It really is a lot simpler and would align more to the way we can do it in the Kotlin Compose template.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -78,6 +100,8 @@ struct NativeLoginView: View {
                     
                     if isAuthenticating {
                         ProgressView()
+                    } else {
+                        Spacer().frame(width: 20, height: 20)
```
JohnsonEricAtSalesforce, line 104: I wasn't sure if `dev` did this, but the progress view was making the layout "overflow" the rectangle background.  So I added a little more room to reserve its space.

wmathurin (changes requested):
Couldn't we pull the RecaptchaEnterprise dependency using cocoapod ??

brandonpage (approved):
LGTM pending Wolf's inquiry about reCAPTCHA being a pod dependency (I was thinking the same thing the other day).

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -155,8 +268,128 @@ struct NativeLoginView: View {
...
+        
+        // Show the progress indicator.
+        isAuthenticating = true
+        
+        // Execute for a new reCAPTCHA token.
+        reCaptchaClientObservable.reCaptchaClient?.execute(withAction: .login) {reCaptchaExecuteResult, error in
+            
+            // Guard for the new reCAPTCHA token.
+            guard let reCaptchaToken = reCaptchaExecuteResult else {
+                print("Could not obtain a reCAPTCHA token due to error with description '\(error?.localizedDescription ?? "(A description wasn't provided.)")'.")
```
bbirman, line 290: for this and other print statements, could SalesforceLogger be used instead?

iOSNativeLoginTemplate/External/ThirdPartyDependencies/RecaptchaEnterprise.xcframework/Info.plist
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
wmathurin: Did we do a 3PP for the reCAPTCHA's lib?

iOSNativeLoginTemplate/Podfile
```diff
@@ -7,11 +7,13 @@ target 'iOSNativeLoginTemplate' do
   source 'https://cdn.cocoapods.org/'
   use_frameworks!
   use_mobile_sdk!
+  
+  pod "RecaptchaEnterprise", "18.5.0-beta01"
```
JohnsonEricAtSalesforce, line 11: This is the version shown in Google's documentation and I'm pretty used to "beta" meaning ready for production in Google-speak.  We could back up to an older stable if anyone see merit.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/AccountsListModel.swift
```diff
@@ -71,7 +71,7 @@ class AccountsListModel: ObservableObject {
                 }
             }
             .catch { error -> Just<[Account]> in
-                print(error)
+                SalesforceLogger.e(AccountsListModel.self, message: error.localizedDescription)
```
JohnsonEricAtSalesforce, line 74: I fixed this one which was also on `dev`.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -155,8 +268,128 @@ struct NativeLoginView: View {
...
+        
+        // Show the progress indicator.
+        isAuthenticating = true
+        
+        // Execute for a new reCAPTCHA token.
+        reCaptchaClientObservable.reCaptchaClient?.execute(withAction: .login) {reCaptchaExecuteResult, error in
+            
+            // Guard for the new reCAPTCHA token.
+            guard let reCaptchaToken = reCaptchaExecuteResult else {
+                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA token due to error with description '\(error?.localizedDescription ?? "(A description wasn't provided.)")'.")
```
JohnsonEricAtSalesforce, line 290: I suppose since this is a struct that's why `NativeLoginView.self` cannot be used here.  It needs a class.  Is there any work-around for that and is it significant?

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
JohnsonEricAtSalesforce: Good suggestion.  Done!  I also cleaned up print use from `dev` as well.

iOSNativeLoginTemplate/External/ThirdPartyDependencies/RecaptchaEnterprise.xcframework/Info.plist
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Not that I'm aware of. What is 3PP (a third-party library review or approval?)

Note: we _could_ remove the library entirely from the template and leave commented instructions where needed.  That might not be the best experience, but I know there are other templates that do that for things such as push notifications.

iOSNativeLoginTemplate/Podfile
brandonpage: Yeah, to them alpha is usually "production quality but we might make API changes".  Beta is pretty much always safe.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
bbirman: Maybe we could add an extension for structs? Out of scope for this but could add it to the backlog, seems like a good modernization item

iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
```diff
@@ -129,3 +163,23 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+            await initializeReCaptchaClient(reCaptchaSiteKey: reCaptchaSiteKey)
+        }
+    }
+    
+    final func initializeReCaptchaClient(reCaptchaSiteKey: String) async {
+        do {
+
+            reCaptchaClient = try await Recaptcha.getClient(withSiteKey: reCaptchaSiteKey)
+        } catch let error {
+            SalesforceLogger.e(AppDelegate.self, message: "Cannot get reCAPTCHA client due to an error with description '\(error.localizedDescription).'.")
```
bbirman, line 182: should this be SceneDelegate or was there another logger issue?

JohnsonEricAtSalesforce: Good catch since the SceneDelegate is a class.  Updated!

bbirman (approved):

wmathurin (approved):

brandonpage (approved):

PR #405, opened 2024-04-20
Setup Native Login for UI testing.

brandonpage:
Because it is very easy to find and replace strings in the UI Testing framework (like `your-client-id`) but difficult to add full LOC from scratch at specific places in file I have made username / password login the default flow.  It is impossible to UI Test Password-less login so its setup, that would otherwise cause the app to not compile, is commented out. 

UI Testing PR: https://github.com/forcedotcom/SalesforceMobileSDK-UITests/pull/39

### DIFF
--- a/iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
+++ b/iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
@@ -53,38 +53,50 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         assert(redirectUri != "your-redirect-uri", "Please add your Native Login redirect uri.")
         assert(loginUrl != "your-community-url", "Please add your Native Login community url.")
         
+        // Used to create a View Controller from SwiftUI.
+        let nativeLoginViewController = NativeLoginViewFactory.create()
+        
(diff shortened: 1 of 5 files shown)


iOSNativeLoginTemplate/iOSNativeLoginTemplate/SceneDelegate.swift
```diff
@@ -53,39 +53,51 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        //
+        // Un-comment the code block below and fill in the values from the Google Cloud project
+        // reCAPTCA settings.  Note that only enterprise reCAPTCHA requires the reCAPTCHA
+        // Site Key Id and Google Cloud Project Id.
+        //
+        // When using non-enterprise reCAPTCHA, set reCAPTCHA Site Key Id and
+        // Google Cloud Project Id to nil along with a false value for the
+        // enterprise parameter.
+        //
+//        let reCaptchaSiteKeyId = "your-recaptcha-site-key-id"
```
bbirman, line 77: not sure if it's worth it here but might want to put all the native login setup in a method and just comment out the helper method call so it's easier to catch any API changes later

bbirman (approved):

JohnsonEricAtSalesforce (approved):
🤘🏻

PR #407, opened 2024-04-25
@W-15611884: [MSDK Android] Android Native Login Template Cannot Be Created Due To Invalid Package Names

JohnsonEricAtSalesforce:
🎸 _Ready For Review_ 🥁

  This is an easy one.  The package names in Kotlin for this template changed after it was created and this catches the template script up with that.

### DIFF
--- a/AndroidNativeLoginTemplate/template.js
+++ b/AndroidNativeLoginTemplate/template.js
@@ -37,7 +37,7 @@ function prepare(config, replaceInFiles, moveFile, removeFile) {
 
     // Values in template
     var templateAppName = 'AndroidNativeLoginTemplate';
-    var templatePackageName = 'com.salesforce.androidnativelogintemplate';
+    var templatePackageName = 'com.salesforce.androidnativelogin';
 
     // Key files
     var templatePackageJsonFile = 'package.json';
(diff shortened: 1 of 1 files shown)


brandonpage:
Where are you seeing a problem with this? `androidnativelogintemplate` seems correct to me.

wmathurin (changes requested):
That does not look right.
I see androidnativelogintemplate in the package path / names e.g. https://github.com/forcedotcom/SalesforceMobileSDK-Templates/tree/dev/AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate

brandonpage:
@JohnsonEricAtSalesforce Did you enter `com.salesforce.androidnativelogin` as the package when you generated the template?  The actual name and package will be whatever you entered when generating it.  If I enter `com.sfdc.test` that will be the applied everywhere.  

However maybe there is an issue with setting the package correctly if it is a substring?  Maybe there are problems replacing `com.salesforce.androidnativelogintemplate` with `com.salesforce.androidnativelogin`?

brandonpage (changes requested):
Okay, I see the problem now but this is the wrong way to solve it.  

There are a bunch of imports in `NativeLogin` that are **hardcoded** references to other files in the app:
```
import com.salesforce.androidnativelogin.MainApplication.Companion.executeLoginAction
import com.salesforce.androidnativelogin.NativeLoginViewModel.IdentityFlowLayoutType
import com.salesforce.androidnativelogin.NativeLoginViewModel.IdentityFlowLayoutType.InitializePasswordLessLoginViaOtp
import com.salesforce.androidnativelogin.NativeLoginViewModel.IdentityFlowLayoutType.LoginViaUsernameAndOtp
import com.salesforce.androidnativelogin.NativeLoginViewModel.IdentityFlowLayoutType.LoginViaUsernamePassword
import com.salesforce.androidnativelogin.R.drawable.radio_button_checked_24px
import com.salesforce.androidnativelogin.R.drawable.radio_button_unchecked_24px
import com.salesforce.androidnativelogin.R.drawable.sf__salesforce_logo
```

If those get changed to `com.salesforce.androidnativelogintemplate` do they get modified by the script to be the user entered package name?

wmathurin:
That line should take care of it: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/AndroidNativeLoginTemplate/template.js#L61

brandonpage:
I believe this is the correct way to fix the issue: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/408

brandonpage:
Fixed by https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/408

PR #416, opened 2024-07-11
@W-15811064: [MSDK Android Templates] Native Login Template Scrolls Registration Text Fields Out Of View On Focus

JohnsonEricAtSalesforce:
🥁 _Ready For Review!_ 🎸

  This corrects the status bar, navigation bar and IME padding for the Android native login template app.  Previously, the taller registration form was scrolling the focused field vertically up and out of view on focus.  Now, all of the layouts will adjust gracefully to these three layout topics.

  I gave @brandonpage a demonstration and a screen-share walk-through as well so he could closely review how this matches with his original layout.

### DIFF
--- a/AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate/NativeLogin.kt
+++ b/AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate/NativeLogin.kt
@@ -35,7 +35,6 @@ import android.widget.Toast
 import android.widget.Toast.LENGTH_LONG
 import android.window.OnBackInvokedDispatcher.PRIORITY_DEFAULT
 import androidx.activity.ComponentActivity
-import androidx.activity.enableEdgeToEdge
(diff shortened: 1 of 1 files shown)


AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate/NativeLogin.kt
```diff
@@ -35,7 +35,6 @@ import android.widget.Toast
 import android.widget.Toast.LENGTH_LONG
 import android.window.OnBackInvokedDispatcher.PRIORITY_DEFAULT
 import androidx.activity.ComponentActivity
-import androidx.activity.enableEdgeToEdge
```
brandonpage, line 38: NIT:  As mentioned before, [apps targeting Android 15 will be edge to edge by default](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge).

brandonpage (approved):

PR #417, opened 2024-07-22
W-16171426: [Android] Update Native Login Sample App to include QR Code Login Flow

JohnsonEricAtSalesforce:
🎸 _Ready For Review!_ 🥁

This is a companion to https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2594.  Do read that in detail before beginning here.

This adds commented code to the Android Native Kotlin template detailing how to enable login via a UI Bridge API generated log in QR code.  This is web login specific, so when I chatted with @wmathurin regarding which template this should live in, we easily ruled out the newer Native Login Template.  It's also an optional feature that builds logically on the Android Native Kotlin Template, so having it commented out in this context seems to make sense without adding too much clutter.

I used a consistent comment phrase throughout to make it easier to document how to find all the touch points in the template.  With those commented out, the template behaves as it does on `dev`.  For anyone wanting to test, just checkout one earlier commit to test with a fully working configuration.


JohnsonEricAtSalesforce:
It's worth noting I did intend to include commits from the spike performed by @wmathurin.  I rebased his branch on `dev` before adding my own commits.  Plus, a squash will make this very tidy on `dev` when we're done while preserving his history here.

AndroidNativeKotlinTemplate/app/build.gradle.kts
```diff
@@ -7,6 +7,9 @@ plugins {
 
 dependencies {
     implementation("com.salesforce.mobilesdk:MobileSync:12.1.0")
+    // Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes
```
JohnsonEricAtSalesforce, line 10: We'll have to see how we want to document using this in the template, but I used the phrase `when enabling log in via Salesforce UI Bridge API generated QR codes` on every touch point.  On a parallel, we need to see how we want to document the companion APEX code that backs this up as well.

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -1,55 +1,47 @@
...
-            'scheme' would be "testsfdc", 'host' would be "*" since it doesn't exist, and
-            'path' would be "/mobilesdk/detect/oauth/done".
-
-            If the callback URL is "sfdc://login.salesforce.com/oauth/done",
-            'scheme' would be "sfdc", 'host' would be "login.salesforce.com",
-            and 'path' would be "/oauth/done".
-        -->
-        <!--
-        <activity android:name="com.salesforce.androidsdk.ui.LoginActivity"
-            android:theme="@style/SalesforceSDK"
```
brandonpage, line 40: I don't think we want to remove this.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
```diff
@@ -0,0 +1,72 @@
+//package com.salesforce.androidnativekotlintemplate
```
brandonpage, line 1: Is there a reason class and the matching xml are commented out?  Would be a lot easier to automate if they were not.

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
JohnsonEricAtSalesforce: That's interesting.  I didn't realize that was also commented out on `dev`.  Should we also bring back the comment block as well? @wmathurin had removed it during his spike.

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -35,21 +36,23 @@
...
             and 'path' would be "/oauth/done".
         -->
-        <!--
-        <activity android:name="com.salesforce.androidsdk.ui.LoginActivity"
-            android:theme="@style/SalesforceSDK"
-            android:launchMode="singleTop"
-            android:exported="true">
+        <!-- Use `android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"` when enabling log in via Salesforce UI Bridge API generated QR codes -->
+<!--        <activity-->
+<!--            android:name="com.salesforce.androidsdk.ui.LoginActivity"-->
```
JohnsonEricAtSalesforce, line 41: @brandonpage - For your previous comment on this section, this is functionally a match for `dev`, but I went with single-line commenting since it uncomments easier in Android Studio with Command-/ and also quotes the inner commented code regarding QR Code Log In.  How's that look?

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
JohnsonEricAtSalesforce: I was wondering here also.  Both of these would only be needed if a user of the template were going to uncomment the QR Code Log In related sources elsewhere.  If that's not done, it seemed nice to omit the need to compile these and have them potentially dangling in the workspace.

Other than running a few extra compilation steps, having them uncommented would be harmless.  Shall we uncomment them?

brandonpage: I was mostly thinking about automating but on second thought that isn't going to be possible so I guess it doesn't make much difference.

AndroidNativeKotlinTemplate/app/build.gradle.kts
```diff
@@ -7,6 +7,9 @@ plugins {
...
+//    implementation("com.google.zxing:core:3.4.1")
+//    implementation("com.journeyapps:zxing-android-embedded:4.3.0")
```
wmathurin, line 12: One drawback with having things to uncomment is that our automated tests can't exercise that code.
Maybe we should have it in the template by default and the comment could say: "Comment the following code if you do no plan to use QR login". Another option is to have a separate template alltogether.

AndroidNativeKotlinTemplate/app/build.gradle.kts
wmathurin: We can probably start by having the apex code in the documentation.
To make it easier we could have the server code in a sfdx project that is part of the template (this one or a new one) so that a dev can simply sfdx push the code to their org. We did that before in the [HybridLwcTemplate](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/tree/dev/HybridLwcTemplate/server).
If we do that, maybe we should have QR login in its own template.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
wmathurin: Even if we can't automate the actual login, test_force could make sure everything compiles.
If it is commented out and any API changes, we won't know unless we manually test (by uncommenting the code).
I recommend we put that code in and have the comment to guide them in case they are not planning to use the feature. Or we could move it to its own template.

AndroidNativeKotlinTemplate/app/build.gradle.kts
JohnsonEricAtSalesforce: We have so many templates, I was considering avoiding having another where the codebase is actually so similar to this one.  They do become toilsome to maintain in great numbers.  That said, this is one of the more complicated customizations we have for the template's user to make - more so than, say, enabling push notifications.

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
JohnsonEricAtSalesforce: @brandonpage - This should be resolved for you in another comment thread.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
JohnsonEricAtSalesforce: Agreed. I uncommented 99.9% of the code, but kept the configuration with QR Code Login disabled via a couple of key lines just like we do for push notifications.

AndroidNativeKotlinTemplate/app/build.gradle.kts
JohnsonEricAtSalesforce: Done!

AndroidNativeKotlinTemplate/app/build.gradle.kts
wmathurin: If we end up having a server folder with a sfdx project that contains an apex page showing a QR code, it will make sense having a new template.

wmathurin (approved):

JohnsonEricAtSalesforce:
Hold merging this to `dev` and the `12.1.0` codebase until we confirm the iOS version will also make it in time for the release.

brandonpage (approved):

PR #426, opened 2024-10-03
@W-16336722: [MSDK Android] QR Code Log In Using External QR Code Reader Does Not Display Main Activity

JohnsonEricAtSalesforce:
🥁 _*Ready For Review!*_ 🎸

This is a companion to https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2625

### DIFF
--- a/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
+++ b/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
@@ -20,6 +20,16 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+
+            <!-- Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes -->
+            <!-- This declares the app's support for QR code login deep links.  The scheme can be customized by the app to match any scheme used in the QR Code's URL -->
+<!--            <intent-filter>-->
+<!--                <data android:scheme="mobileapp" />-->
+<!--                <action android:name="android.intent.action.VIEW" />-->
+
+<!--                <category android:name="android.intent.category.BROWSABLE" />-->
+<!--                <category android:name="android.intent.category.DEFAULT" />-->
+<!--            </intent-filter>-->
         </activity>
 
         <!-- Login activity -->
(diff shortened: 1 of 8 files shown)


AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -20,6 +20,16 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+
+            <!-- Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes -->
+            <!-- This declares the app's support for QR code login deep links.  The scheme can be customized by the app to match any scheme used in the QR Code's URL -->
+<!--            <intent-filter>-->
```
JohnsonEricAtSalesforce, line 26: Note this intent filter is attached to the app's main activity now, which was the crux of fixing this issue.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -156,4 +163,16 @@ class MainActivity : SalesforceActivity() {
...
         })
     }
-}
\ No newline at end of file
+
+    // region QR Code Login Via Salesforce Identity API UI Bridge Public Implementation
+
+    private fun useQrCodeLoginUrl() {
+        val uiBridgeApiParameters = uiBridgeApiParametersFromQrCodeLoginUrl(intent.data.toString())
+        SalesforceSDKManager.getInstance().apply {
```
JohnsonEricAtSalesforce, line 171: Here's the app's opportunity to provide the front door URL before it soon gets the REST client, which will automatically display the login activity as a side effect.

Are there any other paths to displaying the login activity we might need to intercept? Once these parameters are set, the login activity will use them one time.

JohnsonEricAtSalesforce (commented):
Here's my self review and code walk-through to start the discussion.  Thanks!

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
wmathurin: >Are there any other paths to displaying the login activity we might need to intercept? Once these parameters are set, the login activity will use them one time.

I don't think so.

wmathurin (approved):

PR #427, opened 2024-10-08
@W-16814802: [iOS] Add Deep Link From External QR Code Reader To iOS Native Swift Template QR Code Login

JohnsonEricAtSalesforce:
🤘🏻 _*Ready For Review!*_ 🎸

  This is a parallel to https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/426.  The key change for the template is updating the naming conventions, documentation and a few customization options to match the public API for QR Code Log In we just finished for Android.

  Thanks, as always, for reading along ⭐️

### DIFF
--- a/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
+++ b/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
@@ -23,13 +23,13 @@
 
             <!-- Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes -->
             <!-- This declares the app's support for QR code login deep links.  The scheme can be customized by the app to match any scheme used in the QR Code's URL -->
-<!--            <intent-filter>-->
-<!--                <data android:scheme="mobileapp" />-->
-<!--                <action android:name="android.intent.action.VIEW" />-->
+            <intent-filter>
+                <data android:scheme="mobileapp" />
+                <action android:name="android.intent.action.VIEW" />
 
-<!--                <category android:name="android.intent.category.BROWSABLE" />-->
-<!--                <category android:name="android.intent.category.DEFAULT" />-->
-<!--            </intent-filter>-->
+                <category android:name="android.intent.category.BROWSABLE" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
         </activity>
 
         <!-- Login activity -->
@@ -47,22 +47,10 @@
             and 'path' would be "/oauth/done".
         -->
         <!-- Use `android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"` when enabling log in via Salesforce UI Bridge API generated QR codes -->
-<!--        <activity-->
-<!--            android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"-->
-<!--            android:theme="@style/SalesforceSDK"-->
-<!--            android:launchMode="singleTask"-->
-<!--            android:exported="true">-->
-
-<!--            &lt;!&ndash; Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes &ndash;&gt;-->
-<!--            &lt;!&ndash; <intent-filter>&ndash;&gt;-->
-<!--            &lt;!&ndash;     <data&ndash;&gt;-->
-<!--            &lt;!&ndash;         android:host="android"&ndash;&gt;-->
-<!--            &lt;!&ndash;         android:path="/login/qr"&ndash;&gt;-->
-<!--            &lt;!&ndash;         android:scheme="mobileapp" />&ndash;&gt;-->
-<!--            &lt;!&ndash;     <action android:name="android.intent.action.VIEW" />&ndash;&gt;-->
-<!--            &lt;!&ndash;     <category android:name="android.intent.category.BROWSABLE" />&ndash;&gt;-->
-<!--            &lt;!&ndash;     <category android:name="android.intent.category.DEFAULT" />&ndash;&gt;-->
-<!--            &lt;!&ndash; </intent-filter>&ndash;&gt;-->
-<!--        </activity>-->
+        <activity
+            android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"
+            android:exported="true"
+            android:launchMode="singleTask"
+            android:theme="@style/SalesforceSDK" />
     </application>
 </manifest>
--- a/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
+++ b/AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
@@ -23,13 +23,13 @@
 
             <!-- Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes -->
             <!-- This declares the app's support for QR code login deep links.  The scheme can be customized by the app to match any scheme used in the QR Code's URL -->
-            <intent-filter>
-                <data android:scheme="mobileapp" />
-                <action android:name="android.intent.action.VIEW" />
+<!--            <intent-filter>-->
+<!--                <data android:scheme="mobileapp" />-->
+<!--                <action android:name="android.intent.action.VIEW" />-->
 
-                <category android:name="android.intent.category.BROWSABLE" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
+<!--                <category android:name="android.intent.category.BROWSABLE" />-->
+<!--                <category android:name="android.intent.category.DEFAULT" />-->
+<!--            </intent-filter>-->
         </activity>
 
         <!-- Login activity -->
@@ -47,10 +47,10 @@
             and 'path' would be "/oauth/done".
         -->
         <!-- Use `android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"` when enabling log in via Salesforce UI Bridge API generated QR codes -->
-        <activity
-            android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"
-            android:exported="true"
-            android:launchMode="singleTask"
-            android:theme="@style/SalesforceSDK" />
+<!--        <activity-->
+<!--            android:name="com.salesforce.androidsdk.ui.LoginActivity"-->
+<!--            android:exported="true"-->
+<!--            android:launchMode="singleTask"-->
+<!--            android:theme="@style/SalesforceSDK" />-->
     </application>
 </manifest>
--- a/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
+++ b/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
@@ -26,6 +26,9 @@
  */
 package com.salesforce.androidnativekotlintemplate
 
+import android.content.Intent
+import android.content.Intent.FLAG_ACTIVITY_SINGLE_TOP
+import android.net.Uri
 import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES.UPSIDE_DOWN_CAKE
 import android.os.Bundle
@@ -45,9 +48,14 @@ import com.salesforce.androidsdk.rest.RestClient
 import com.salesforce.androidsdk.rest.RestClient.AsyncRequestCallback
 import com.salesforce.androidsdk.rest.RestRequest
 import com.salesforce.androidsdk.rest.RestResponse
-import com.salesforce.androidsdk.ui.LoginActivity.Companion.isQrCodeLoginIntent
-import com.salesforce.androidsdk.ui.LoginActivity.Companion.uiBridgeApiParametersFromQrCodeLoginUrl
+import com.salesforce.androidsdk.ui.LoginActivity
+import com.salesforce.androidsdk.ui.LoginActivity.Companion.EXTRA_KEY_FRONTDOOR_BRIDGE_URL
+import com.salesforce.androidsdk.ui.LoginActivity.Companion.EXTRA_KEY_PKCE_CODE_VERIFIER
+import com.salesforce.androidsdk.ui.LoginActivity.Companion.isQrCodeLoginUrlIntent
+import com.salesforce.androidsdk.ui.LoginActivity.Companion.qrCodeLoginUrlJsonParameterName
+import com.salesforce.androidsdk.ui.LoginActivity.Companion.qrCodeLoginUrlPath
 import com.salesforce.androidsdk.ui.SalesforceActivity
+import com.salesforce.androidsdk.util.SalesforceSDKLogger.e
 import java.io.UnsupportedEncodingException
 import java.util.*
 
@@ -61,11 +69,6 @@ class MainActivity : SalesforceActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-
-        // Check for and use the intent's QR code login URL if applicable.
-        if (isQrCodeLoginIntent(intent)) {
-            useQrCodeLoginUrl()
-        }
 
         // Set Theme
         val idDarkTheme = MobileSyncSDKManager.getInstance().isDarkTheme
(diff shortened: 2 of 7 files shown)


iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/SceneDelegate.swift
```diff
@@ -109,4 +139,34 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        // Use login QR code if applicable
+        
+        // Validate the URL is a valid QR code login URL.
+        guard let components = NSURLComponents(
+            url: url,
+            resolvingAgainstBaseURL: true),
+              let scheme = components.scheme,
+              scheme == "mobileapp",
+              let host = components.host,
+              host == "android",
```
bbirman, line 161: are we using android as the example for both platforms?

iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/SceneDelegate.swift
```diff
@@ -109,4 +139,34 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+              let path = components.path,
+              path == SalesforceLoginViewController.qrCodeLoginUrlPath,
+              let queryItems = components.queryItems,
+              let _ = queryItems.first(where: { $0.name == SalesforceLoginViewController.qrCodeLoginUrlJsonParameterName })?.value else {
+            print("Invalid QR code log in URL.")
```
bbirman, line 166: nit: can print be swapped for SalesforceLogger.e?

iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/SceneDelegate.swift
JohnsonEricAtSalesforce: I pondered changing it here, even though the actual string value in the template is arbitrary and should be changed by the template user.  If we did want to change it, we could follow-up with pull requests for both Android and iOS to pick a common value and then ensure the reference Apex/VIsualforce Page match.

What if we pulled a page from our other recent authentication efforts and put in a placeholder value that is followed by an assertion requiring it have been changed?

JohnsonEricAtSalesforce: @bbirman - This looks like a winner, though it highlights that setting up QR code log in is bit complicated. 
https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/427/commits/7833e403f7eeba12f3e92539a406c86c59c2905b

@brandonpage & @wmathurin, if this looks good on the iOS side I reckon a quick commit to the Android template to match would be really helpful.

wmathurin: 👍 we should use a place holder value, assert it get replaced, document it needs to match the server side.
In 12.2, the apex page will be in the doc, right?

JohnsonEricAtSalesforce: Correct on both ☝🏻

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -48,21 +48,9 @@
...
         <!-- Use `android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"` when enabling log in via Salesforce UI Bridge API generated QR codes -->
 <!--        <activity-->
-<!--            android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"-->
-<!--            android:theme="@style/SalesforceSDK"-->
+<!--            android:name="com.salesforce.androidsdk.ui.LoginActivity"-->
+<!--            android:exported="true"-->
 <!--            android:launchMode="singleTask"-->
-<!--            android:exported="true">-->
-
-<!--            &lt;!&ndash; Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes &ndash;&gt;-->
```
JohnsonEricAtSalesforce, line 56: This was still commented out here even though it also is commented on main activity where it belongs.

AndroidNativeKotlinTemplate/app/src/main/AndroidManifest.xml
```diff
@@ -48,21 +48,9 @@
         -->
         <!-- Use `android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"` when enabling log in via Salesforce UI Bridge API generated QR codes -->
 <!--        <activity-->
-<!--            android:name="com.salesforce.androidnativekotlintemplate.QrCodeEnabledLoginActivity"-->
-<!--            android:theme="@style/SalesforceSDK"-->
+<!--            android:name="com.salesforce.androidsdk.ui.LoginActivity"-->
```
JohnsonEricAtSalesforce, line 51: This should be LoginActivity since this block is for the browser based auth comment flow.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -62,11 +70,6 @@ class MainActivity : SalesforceActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        // Check for and use the intent's QR code login URL if applicable.
-        if (isQrCodeLoginIntent(intent)) {
```
JohnsonEricAtSalesforce, line 66: Having this check in on-create was fine for just deep-linked QR codes, but it works better in on-resume for other cases such as returning from the QR code scan activity.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -90,21 +93,38 @@ class MainActivity : SalesforceActivity() {
...
+            /*
+             * Temporarily request the Salesforce Mobile SDK REST Client not be
+             * built on resume such that the default login activity will not
+             * display and conflict with the login activity for QR code log in.
+             *
+             * If the display of the default login activity was no longer a
+             * side-effect of getting the REST client, this variable could be
+             * removed.
+             */
+            buildRestClientOnResume = false
```
JohnsonEricAtSalesforce, line 114: Much like a similar comment on https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2628, the app's ability to receive data in main activity is greatly limited by the automatic display of the default log in activity.  This returns some control to the app until possibly some of that tightly coupled logic can be refactored.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -184,12 +204,61 @@ class MainActivity : SalesforceActivity() {
...
-    private fun useQrCodeLoginUrl() {
-        val uiBridgeApiParameters = uiBridgeApiParametersFromQrCodeLoginUrl(intent.data.toString())
-        SalesforceSDKManager.getInstance().apply {
-            frontDoorBridgeUrl = uiBridgeApiParameters?.frontdoorBridgeUrl
-            frontDoorBridgeCodeVerifier = uiBridgeApiParameters?.pkceCodeVerifier
+    /**
+     * Validates and uses the intent's QR code log in URL.
+     * @param url The QR code log in URL
+     */
+    private fun useQrCodeLogInUrl(url: Uri) {
```
JohnsonEricAtSalesforce, line 211: This method is now the app's single path for QR Code Log In from external deep links _and_ the internal QR code scanner.  Half that logic used to live in the activity with the QR code scanner.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainApplication.kt
```diff
@@ -67,8 +74,92 @@ class MainApplication : Application() {
...
 
-        // Enable login via Salesforce UI Bridge API generated QR code.
-        // Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes
-        // MobileSyncSDKManager.getInstance().isQrCodeLoginEnabled = true
+        /* Uncomment when enabling log in via Salesforce UI Bridge API generated QR codes. */
+        //setupQrCodeLogin()
+    }
+
+    // endregion
+    // region QR Code Login Via Salesforce Identity API UI Bridge Public Implementation
```
JohnsonEricAtSalesforce, line 82: The documentation and API here now match the iOS side exactly 🌈

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
```diff
@@ -35,6 +43,29 @@ class QrCodeEnabledLoginActivity : LoginActivity() {
...
+                 * Note: When this activity starts the intent for main activity
+                 * after scanning the QR code, finishing prior to log in
+                 * completion would also finish the activities being used for
+                 * the in-progress log in.
+                 *
+                 * If the main activity intent could be started in such a way
+                 * that this activity could finish immediately after starting
+                 * the intent, this receiver wouldn't be needed.
+                 */
+                if (intent.action == USER_SWITCH_INTENT_ACTION) finish()
```
JohnsonEricAtSalesforce, line 60: @brandonpage or @wmathurin, here's a question that should be easy to solve, but for some reason I was stumped.

A little later on, this activity wants to start an intent for main activity with the QR Code Log In data.  That's fine, except if this intent finishes before the user is authenticated, it finishes the entire app back to the launcher.

I tried quite a few different permutations of setting up the intent and activities, but to no avail.  Since it finishes with no consequence _after_ the user is authenticated I'm reckoning there's logic inherited from LoginActivity that is up to something.  With this work-around, perhaps that's not a release blocker, but I really do just want to know what it is.  🧐

iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/AppDelegate.swift
```diff
@@ -116,4 +109,96 @@ class AppDelegate : UIResponder, UIApplicationDelegate {
...
+    // MARK: - QR Code Login Via Salesforce Identity API UI Bridge Public Implementation
+    
+    /**
+     * When enabling log in via Salesforce UI Bridge API generated QR codes, choose to use the
+     * Salesforce Mobile SDK reference format for QR code log in URLs or an entirely custom format.
+     *
+     * If only one of the two formats is used, which is the most likely implementation for apps using this
+     * template, this variable and code related to the unused implementation could be removed.
+     */
+    let isQrCodeLoginUsingReferenceUrlFormat = true
```
JohnsonEricAtSalesforce, line 122: @bbirman - I moved this from the SceneDelegate so that the documentation doesn't have the template user updating more files than needed.

JohnsonEricAtSalesforce (commented):
Here's one (final!) round of updates to make Android and iOS have comparable documentation for QR Code Login plus a slight cleanup on iOS by moving everything into the AppDelegate.

wmathurin (changes requested):
NIT: incorrect use of login (noun) vs log in (verb) in comments
I think:
- it should be "QR code login URL"  rather than "QR code log in URL".
- it should be "enabling login via ..." rather than "enabling log in via"

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
brandonpage: I don't think I have a quick alternative but I really don't like this.  This feels like something the SDK should be handling internally

wmathurin: Could we bury that flag deeper maybe between SalesforceSDKMgr and SalesforceActivityDelegate ??

brandonpage: Maybe QR Code login needs it's own activity?  We don't have this problem with Native Login because that is the activity that gets launched when on resume happens with no users logged in.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
brandonpage: Seems similar to this situation maybe? https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/c998f718172710896e724706875f5e4a28a7c891/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt#L228

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
JohnsonEricAtSalesforce: I’m glad we mentioned handling this “internally to the SDK”. The trouble here is that SalesforceActivity does attempt to handle this internally - with no option to make exceptions.  Before we go the route of creating a “main” activity that doesn’t inherit from SalesforceActivity (A suggestion from @Brandonpage I really considered - it makes sense, though hear me out), I have a thought on how we might avoid some redundant code and maybe make SalesforceActivity a more re-usable class for the future. 

The goal is passing the QR Code Login URL/Frontdoor URL/Code Verifier from MainActivity to our existing LoginActivity so we get gobs of code re-use.  I’m sensing we need a loosely coupled way to do things like this in general, but our big singleton is the opposite of loosely coupled (Ain’t it a stinker?). I’d love to avoid adding more properties to that crazy thing.

The trouble is SalesforceActivity is hard-coded to always start LoginActivity in just one way and the app doesn’t have an API to change that.  `onResume` thinks it has to start LoginActivity every time and since that call is buried in ClientManager a few levels down the stack it’s not customizable.

The funny thing is SalesforceActivityDelegate has a `public void onResume(boolean buildRestClient)` and it’s really the `buildRestClient` the app would like to set to `false`.  That’s been a public method for seven years, but the delegate itself is private.  Can we give that to the apps? We simply need a means for the app to use this existing and amazingly public parameter.  Maybe we just need a great name and doc comment so the semantic makes sense?

For now, is there a way to clean up this new buildRestClientOnResume property so it makes sense, maybe even in a non-deprecated way?

This is just my thinking, but ideal future state is we don’t force a specific LoginActivity in on-resume.  If LoginActivity provides some app-consumable options the app could have the API to start LoginActivity itself.  That’s where we need an API allowing the app to opt-out of the automatic, default LoginActivity - When it wants to.

No after hours reply needed 💤 Thanks for pondering a great team reply on this little stumper.

JohnsonEricAtSalesforce: Sleeping on it, I tried out cleaning up the new property this morning so it's benefits and usage are much more clear.

👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/427/commits/dd345a384bee8bbe69b2493171e102dd27a28afc

As I thought about my own experience developing QR code log in as an MSDK app developer, I really thought there was something missing without this property.  The concern I see myself having, in the app developer persona, is that I could use this template and SalesforceActivity to bootstrap my new app and all would seem great - at first.  Then, after investing in building on SalesforceActivity I realize later there's no way to customize the activity's strict on-resume flow (The runtime enforces the call to `super.onResume()`).  It'd be tough, then, to have to retrofit off SalesforceActivity and onto an entirely custom one.

Since MSDK already has a property we can present by-proxy to the app developer we can open up SalesforceActivity to a new scale and provide an off-ramp from the hard-coded on-resume flow - only when needed.

Thoughts? We _could_ go back to adding the new properties on SalesforceSDKManager singleton, though I feel we have a nice opportunity to modernize this activity and avoid building on our old pattern.

JohnsonEricAtSalesforce: Just for reference on how odd it looked using the singleton variables on SalesforceSDKManager, that's here 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-Templates/commit/e5980e4c418bd73e8f1cde5d50b8d0786c4e0957#diff-dfb6fb02b5c24349900f92938b21c03c85aad96aa277a2c36ec2886a51696c80R189

JohnsonEricAtSalesforce:
> NIT: incorrect use of login (noun) vs log in (verb) in comments I think:
> 
> * it should be "QR code login URL"  rather than "QR code log in URL".
> * it should be "enabling login via ..." rather than "enabling log in via"

Yeah, I believe that's a grammar lesson I finally get now 🤣

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
JohnsonEricAtSalesforce: I'm not sure that's the same issue, but I _did_ find the root cause.  `LoginActivity` has some logic where it's making some private choices about the tasks and back stack for us.  In this case, there's a trigger where it's sending the task with it and `MainActivity` to the back too soon.

https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/c998f718172710896e724706875f5e4a28a7c891/libs/SalesforceSDK/src/com/salesforce/androidsdk/ui/LoginActivity.kt#L432

The interesting thing is I'm trying to test all the cases the comment mentioned - they all seem to work just as well with only `finish()` instead of `moveTaskToBack(true)`.  In this event we're actually coming in from `onDestroy()` since the activity is already finished.

I did some research into this logic and it's been in place since this was Java in 2011.
Screenshot 2024-10-10 at 12 25 04

@brandonpage or @wmathurin - With all the changes that have been made over the years to the way `LoginActivity` is placed on the task and back stack, is this even needed?

To minimize risk, we could certainly keep this work-around in the template for a moment.  Then we could analyze this in more detail for a fix or fully understanding it.

brandonpage: @JohnsonEricAtSalesforce It's possible that it is no longer needed on newer devices but older devices may act differently.  For example if you tested on Android 13+ it would use the completely new [predictive back system](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2617/files).  I would also be cautious because LoginActivty is sometimes a lock screen (Bio Auth) and the behavior around custom tabs can further complicate things since it launchers out LoginActivity.

wmathurin (approved):

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
JohnsonEricAtSalesforce: Thanks for the feedback plus the history, @brandonpage. I'm thinking we minimize risk for now and not change this.  In light of the launchMode changes maybe we can revisit this as a maintenance topic soon after the next release.

PR #431, opened 2024-10-16
[MobileSyncExplorerSwift] Fix for tabs hiding nav bar buttons on iOS 18 iPads

bbirman:
Updated so that the nav bar buttons will be positioned to the left on iPads to avoid the overlapping issue we started seeing on iOS 18.

Screenshot 2024-10-15 at 10 39 16 PM
Screenshot 2024-10-15 at 11 07 21 PM
Screenshot 2024-10-15 at 11 09 49 PM


brandonpage (approved):

wmathurin (commented):
👍

PR #436, opened 2024-11-07
Fix React Native Android First Time Login Crash.

brandonpage:
This was already done for the Android MobileSync React Native template with #432, but I discovered today it is needed for all templates.  

No need to patch or anything since this only affects debug builds.


JohnsonEricAtSalesforce (approved):
👍🏻

wmathurin (approved):

PR #442, opened 2025-01-31
@W-17433472: Update Templates

JohnsonEricAtSalesforce:
🎸 _*Ready For Full Review!!!*_ 🥁

  This updates the Android templates where needed with MSDK support for Jetpack Compose.


AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/QrCodeEnabledLoginActivity.kt
```diff
@@ -66,18 +62,24 @@ class QrCodeEnabledLoginActivity : LoginActivity() {
             IntentFilter(USER_SWITCH_INTENT_ACTION),
             ContextCompat.RECEIVER_NOT_EXPORTED
         )
+
+        // Add the Log In With QR Code additional button.
+        viewModel.additionalBottomBarButtons.run {
+            val buttons = value.toMutableList()
+            buttons.add(
```
JohnsonEricAtSalesforce, line 69: @wmathurin, @brandonpage, I put this one up as a really draft pull request so we could look at this.  Here's a new `LoginViewModel` property that we or anyone could use to add additional bottom bar buttons.  This could also work for IDP Auth and Biometric Auth as well, I reckon.  @brandonpage, this also does away with the need for a `LoginViewModel` in this case.

wmathurin: Interesting. I've added the IDP button back [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2658/files#diff-321c03965ac0663f16817520073fbe2d6def698c5568ab388f1161c35536d069R172) but that could be reworked.

wmathurin (approved):

PR #451, opened 2025-05-08
@W-18250527: Websockets client (Android)

JohnsonEricAtSalesforce:
🚨 _*Do Not Merge: Only For Use In Code Review*_ 🚨

### DIFF
--- a/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
+++ b/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
@@ -30,25 +30,32 @@ import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_SINGLE_TOP
 import android.net.Uri
 import android.os.Build.VERSION.SDK_INT
+import android.os.Build.VERSION_CODES.TIRAMISU
 import android.os.Build.VERSION_CODES.UPSIDE_DOWN_CAKE
 import android.os.Bundle
+import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ArrayAdapter
 import android.widget.ListView
 import android.widget.Toast
 import androidx.activity.enableEdgeToEdge
+import androidx.annotation.RequiresApi
 import androidx.core.view.ViewCompat.setOnApplyWindowInsetsListener
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.WindowInsetsCompat.Type.displayCutout
 import androidx.core.view.WindowInsetsCompat.Type.systemBars
 import androidx.core.view.updatePadding
 import com.salesforce.androidnativekotlintemplate.R.id.root
+import com.salesforce.androidnativekotlintemplate.R.raw.asf_api_client_transcription_demo_input
 import com.salesforce.androidsdk.app.SalesforceSDKManager
+import com.salesforce.androidsdk.auth.HttpAccess.DEFAULT
 import com.salesforce.androidsdk.mobilesync.app.MobileSyncSDKManager
+import com.salesforce.androidsdk.rest.AgentforceSpeechFoundationsApiClient
 import com.salesforce.androidsdk.rest.ApiVersionStrings
 import com.salesforce.androidsdk.rest.RestClient
 import com.salesforce.androidsdk.rest.RestClient.AsyncRequestCallback
+import com.salesforce.androidsdk.rest.RestClient.ClientInfo
 import com.salesforce.androidsdk.rest.RestRequest
 import com.salesforce.androidsdk.rest.RestResponse
 import com.salesforce.androidsdk.ui.LoginActivity
@@ -59,8 +66,26 @@ import com.salesforce.androidsdk.ui.LoginActivity.Companion.qrCodeLoginUrlJsonPa
 import com.salesforce.androidsdk.ui.LoginActivity.Companion.qrCodeLoginUrlPath
 import com.salesforce.androidsdk.ui.SalesforceActivity
 import com.salesforce.androidsdk.util.SalesforceSDKLogger.e
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers.Default
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import okhttp3.Response
+import okhttp3.WebSocket
+import okhttp3.WebSocketListener
+import okio.ByteString
+import okio.ByteString.Companion.toByteString
+import java.io.InputStream
 import java.io.UnsupportedEncodingException
+import java.lang.System.arraycopy
+import java.nio.ByteBuffer.allocate
+import java.nio.ByteBuffer.wrap
 import java.util.*
+import java.util.Arrays.copyOfRange
+import kotlin.math.min
+import kotlin.text.Charsets.UTF_8
 
 /**
  * Main activity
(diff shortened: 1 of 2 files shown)


AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -110,12 +135,52 @@ class MainActivity : SalesforceActivity() {
...
         // Keeping reference to rest client
         this.client = client
 
         // Show everything
         findViewById<ViewGroup>(root).visibility = View.VISIBLE
+
+        // Launch to ASF API client demo.
+        client?.let { restClientUnwrapped ->
+            // Use a custom REST client, optionally with a custom access token and access token refresh disabled.
+            val restClientCustomAccessToken = RestClient(
```
JohnsonEricAtSalesforce, line 149: This shows how we can use the existing `RestClient` constructor to create a `RestClient` that provides a custom access token and disables token refresh.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -110,12 +135,52 @@ class MainActivity : SalesforceActivity() {
...
+                restClientUnwrapped.clientInfo.run /* Copy */ {
+                    ClientInfo(
+                        instanceUrl,
+                        loginUrl,
+                        identityUrl,
+                        accountName,
+                        username,
+                        userId,
+                        orgId + "_custom_access_token", // TODO: Inquire if there's another way to circumvent the cached OAuth interceptors in RestClient.setOAuthRefreshInterceptor(authToken). ECJ20250507
```
JohnsonEricAtSalesforce, line 158: @wmathurin - Here's a wrinkle.  The comment says it all.  There's "black magic" in the `RestClient` constructor's call to `setOAuthRefreshInterceptor` that will bypass the value provided here with a cached interceptor.  It's keyed on `userId` and `orgId`.  Overriding one of the key values here is seemingly harmless since the `authTokenProvider` value is also null and `orgId` was only used in the now-disabled token refresh flow.

We could maybe add a dedicated parameter to the `RestClient` constructor that changes the caching behavior.  That would be cleaner, semantically, but I wanted to ask first.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -260,5 +325,124 @@ class MainActivity : SalesforceActivity() {
...
+    /**
+     * Starts the Agentforce Speech Foundations API client demo.
+     */
+    @RequiresApi(TIRAMISU)
+    private fun startAsfApiClientTranscriptionsDemo(
+        restClient: RestClient
+    ) = CoroutineScope(Default).launch {
+
+        // Open the web socket.
+        val webSocket = AgentforceSpeechFoundationsApiClient(
```
JohnsonEricAtSalesforce, line 340: This is a working demonstration ✅ It does as advertised and actually fetches the text for the audio to the logs.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -260,5 +324,129 @@ class MainActivity : SalesforceActivity() {
...
+    /**
+     * Starts the Agentforce Speech Foundations API client demo.
+     */
+    @RequiresApi(TIRAMISU)
+    private fun startAsfApiClientTranscriptionsDemo(
+        restClient: RestClient
+    ) = CoroutineScope(Default).launch {
+
+        Log.i("AsfApiClientTranscriptionsDemo", "Waiting 61 seconds for JWT expiration before opening web socket...")
+        delay(61000)
```
JohnsonEricAtSalesforce, line 339: Note: I added this delay to test that our authorization expiration detection, refresh and retry works for the web socket request.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
```diff
@@ -260,5 +324,129 @@ class MainActivity : SalesforceActivity() {
...
+        while (chunkStartOffset < bytes.size) {
+            val chunkEndOffset = min(
+                (chunkStartOffset + chunkSize).toDouble(),
+                bytes.size.toDouble()
+            ).toInt()
+            val chunk = copyOfRange(bytes, chunkStartOffset, chunkEndOffset)
+            val fullSizedChunk = ByteArray(chunkSize)
+            arraycopy(chunk, 0, fullSizedChunk, 0, chunk.size)
+
+            Log.i("AsfApiClientTranscriptionsDemo", "Sending audio chunk $chunkStartOffset/$chunkEndOffset of ${bytes.size.toDouble()} then a 3 second delay to wait for JWT expiration...")
```
JohnsonEricAtSalesforce, line 374: I also added more delays to the chunk sending to see if the web socket was affected by the JWT expiration while it was open.  It does not seem to be.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
wmathurin: You are right, we need a way to provide an alternate access token. Also the existing refresh logic should not kick in if that access token expires (or we should offer a way to plug in alternate refresh logic as well).

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
wmathurin: So for this test to run, test_credentials need to point to an org with JWT-based access token enabled and set to expire after 1 minute?

wmathurin: Nevermind, we are in the templates repo.

AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
JohnsonEricAtSalesforce: Unpacking this in a bit more detail:
1. The `RestClient` constructor does allow an app to provide the alternate access token (Internally, this is given to a fixed `OAuthRefreshInterceptor` type for things like `buildAuthenticatedRequest`)
2. The app can disable the existing refresh logic by passing `null` for the `authTokenProvider` parameter of the `RestClient` constructor
3. We don't have a way for the app to provide their own refresh logic.  Would we want to go a little more abstract and maybe let the app provide their own instance of a "RestClient" entirely? I don't believe we have a way for apps to inject a `RestClient` implementation.  Or, maybe we let the app pass their own implementation of `OAuthRefreshInterceptor`?

That's a lot of detail.  I'm thinking out loud a bit.  We also have to potentially disable caching for a RestClient's `OAuthRefreshInterceptor` in `setOAuthRefreshInterceptor` as well.

Moving forward, what if:
1. We let the app provide a custom instance of `OAuthRefreshInterceptor` in the `RestClient` constructor to provide custom authorization expiration detection, re-authorization and retry logic
2. We also add a `RestClient` constructor parameter to disable the `OAuthRefreshInterceptor` Maybe this could be implicit if a custom `OAuthRefreshInterceptor` is provided to the constructor?

JohnsonEricAtSalesforce:
This pull request may have useful code for the future, but for now it was only for example purposes in reviewing the MSDK changes in https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2712

PR #457, opened 2025-07-10
Create a native iOS Template using agentforce.

asikora-sf:
Builds on top of the existing iOS template with dependencies on AgentforceSDK and AgentforceService.


wmathurin:
@asikora-sf We feel that this template should be hosted in a repo you own so that you have full control in terms of releases / patches etc. We can get on a call to chat more.

PR #497, opened 2026-04-17
Temporary fix for React Native black screen issue on Android API 32, 33 and 34.

brandonpage:


JohnsonEricAtSalesforce (approved):
👍🏻

MobileSyncExplorerReactNative/android/app/src/main/java/com/salesforce/samples/mobilesyncexplorerreactnative/MainActivity.kt
```diff
@@ -59,4 +71,29 @@ class MainActivity : SalesforceReactActivity() {
...
+     * transition from "not logged in" to "logged in" after a pause (i.e. the
+     * user just completed OAuth on top of this activity), we force a clean
+     * activity recreate so React Native mounts fresh with the authenticated
+     * user and a properly measured surface.  Subsequent setRestClient() calls
+     * (e.g. on the recreated activity, or returning from background) take the
+     * "already logged in" branch and do not recreate, so there is no loop.
+     *
+     * TODO: Remove this workaround when the issue is fixed in Mobile SDK 14.0.
+     */
+    override fun setRestClient(restClient: RestClient?) {
```
wmathurin, line 90: Should we do that in SalesforceReactActivity (for all apps to benefit from it)?
How would it work with deferred login scenario? Would it interfere with that logic: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2863

PR #498, opened 2026-04-30
Organize .claude/skills for SDK developers vs SDK consumers

wmathurin:
- Mark all existing skills as internal (metadata.internal: true) since they are for SDK contributors, not app developers using the SDK
- Convert update-ios-deployment-target from directory/SKILL.md to flat .md format, consistent with other internal skills
- Add .claude/skills/README.md explaining the two-audience convention: internal skills use flat .md files (local Claude Code use only), consumer skills use subdirectory/SKILL.md format (discoverable via npx skills add forcedotcom/SalesforceMobileSDK-Templates)

### DIFF
--- a/.claude/skills/README.md
+++ b/.claude/skills/README.md
@@ -1,0 +1,72 @@
+# Mobile SDK Skills
+
+This directory contains agent skills for [Claude Code](https://code.claude.com/docs/en/skills) and other compatible AI coding agents.
+
+Skills are installable via the [vercel-labs/skills](https://github.com/vercel-labs/skills) CLI:
+
+```bash
+npx skills add forcedotcom/SalesforceMobileSDK-Templates
+```
+
+## Two Audiences, One Directory
+
+Skills are split into two groups using the `metadata.internal` flag.
+
+### Consumer skills (default)
+
+Skills intended for **developers building apps with Mobile SDK** (SDK consumers). These are visible by default when installing via the CLI above.
+
+Use the directory/`SKILL.md` format so the `npx skills` CLI can discover and install them:
+
+```
+.claude/skills/
+  my-consumer-skill/
+    SKILL.md
+```
+
+_No consumer skills yet — add them here._
+
+### SDK developer skills (internal)
+
+Skills intended for **developers working on Mobile SDK itself** (contributors, maintainers). These are hidden from the default skill listing and only loaded by Claude Code when the repo is open locally — the `npx skills` CLI is not the delivery mechanism.
+
+Because they are only ever used via local repo clone, a flat `.md` file is sufficient (no need for a subdirectory):
+
+```
+.claude/skills/
+  my-sdk-skill.md
+```
+
+| Skill | Description |
+|---|---|
+| `remove-template.md` | Remove a template from this repository |
+| `test-template.md` | Test templates with `test_template.sh` |
+| `update-ios-deployment-target.md` | Bump the minimum iOS deployment target across all templates |
+
+## Adding a New Skill
+
+**Consumer skill** — create a subdirectory with a `SKILL.md`:
+
+```markdown
+---
+name: my-skill
+description: What this skill does and when to use it
+---
+
+# My Skill
+...
+```
+
+**SDK developer skill** — create a flat `.md` file with `metadata.internal: true`:
+
+```markdown
+---
+name: my-skill
+description: What this skill does and when to use it
+metadata:
+  internal: true
+---
+
+# My Skill
+...
+```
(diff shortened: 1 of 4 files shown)


.claude/skills/README.md
```diff
@@ -0,0 +1,72 @@
+# Mobile SDK Skills
+
+This directory contains agent skills for [Claude Code](https://code.claude.com/docs/en/skills) and other compatible AI coding agents.
```
brandonpage, line 3: ~~I don't think we should reference Claude.  [Skills are an open standard](https://agentskills.io/home).~~

Edit: Very, NIT.

.claude/skills/README.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+**SDK developer skill** — create a flat `.md` file with `metadata.internal: true`:
+
+```markdown
+---
+name: my-skill
+description: What this skill does and when to use it
+metadata:
+  internal: true
```
sfdctaka, line 67: Leaving a note here in case anyone needs to know what `metadata:internal:true` does as I had to look it up myself: 
```
In a SKILL.md file, metadata: internal: true hides an AI agent skill from automatic discovery or public listing. It is primarily used for work-in-progress, experimental, or private tooling skills that should not be visible to users during normal skill browsing or installation.
```

.claude/skills/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+.claude/skills/
+  my-consumer-skill/
+    SKILL.md
+```
+
+_No consumer skills yet — add them here._
+
+### SDK developer skills (internal)
```
sfdctaka, line 29: Just curious, is there always going to be a clear demarcation between the consumer and SDK developer?

sfdctaka (approved):
LGTM!

brandonpage (approved):

.claude/skills/README.md
brandonpage: It sounds like we will just have 1 consumer MSDK skill that encompass everything so hopefully it won't feel disorganized/cluttered at the top level.

.claude/skills/README.md
brandonpage: Also should be noted that this is specific to `vercel` and not part of the official spec (AFIK).

.claude/skills/README.md
wmathurin: I think there is a clear demarcation:
- skills meant for making changes to the templates repo (e.g. adding a new template, changing the iOS min version etc) are SDK devs skills (only used by us or anyone contributing to the templates repo),
- skills meant to be used by app developers that consume Mobile SDK and its templates and probably published elsewhere by MPE are SDK consumer skills.

We don't want MPE to pick up the SDK devs skills. It won't make any sense for instance for an app developer to want to change the SQLCipher version or min iOS version that a given template uses - because that's constrained by what the underlying Mobile SDK supports.

.claude/skills/README.md
wmathurin: Exactly, if someone does a `npx skills add https://github.com/forcedotcom/SalesforceMobileSDK-Templates`, we don't want them to install skills meant for doing work on the templates repo. We also don't want MPE to copy those internal skills out to whatever repo they will use to aggregate various Salesforce mobile skills.

bbirman:
The claude `/skills` command doesn't show the skill for me with the flat format

wmathurin:
@bbirman even though the SDK devs skills are not visible using /skills, you can still use them — but you need to invoke them by name directly rather than picking from the dialog. Type /remove-template, /test-template, or /update-ios-deployment-target in the prompt and Claude Code will find and run them.  

The /skills dialog is a discovery UI — it lists skills available to choose from. The metadata.internal: true flag hides skills from that listing (which is what we wanted, to keep the consumer-facing list clean). But the skills are still present on disk and Claude Code can still load and execute them when invoked by name.  

One thing worth knowing: this behaviour — Claude Code respecting metadata.internal in the dialog — is a bonus we got for free. The flag was designed for the npx skills CLI, but Claude Code apparently honours it in the same way.

bbirman:
Hmm, how are you doing it?

I get `Unknown command: /update-ios-deployment-target`

wmathurin:
> Hmm, how are you doing it?
> 
> I get `Unknown command: /update-ios-deployment-target`

@bbirman It's fixed now - see https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/499/changes/68cd15f2a07eaf9ef58dd69e0aea07cecde3f6b4
They show up in `/skills` and are found also directly for instance with `/update-ios-deployment-target`.

PR #501, opened 2026-05-04
Add actionable push notification support to iOSNativeSwiftEncryptedNotificationTemplate

sfdctaka:
Summary                                                                                               
                                                                                                        
  - Adds UNUserNotificationCenterDelegate conformance to AppDelegate to handle foreground notification  
  presentation and action button taps
  - Calls fetchAndStoreNotificationTypes() after Salesforce push registration and on each app foreground
   to keep action button definitions current                                                            
  - Calls invokeServerNotificationAction when the user taps an action button, passing the nid from the
  sfdc payload dictionary (not the top level)                                                           
  - Fixes template.js to include NotificationServiceExtension.entitlements in app name substitution —
  without this, the extension's keychain access group would retain the template name after generation,  
  breaking encrypted notification decryption
                                                                                                        
  Test plan                                                                                             
   
  - Generate an app with forceios createwithtemplate using this template                                
  - Verify NotificationServiceExtension.entitlements contains the generated app name (not
  EncryptedNotificationTemplate)                                                                        
  - Deploy to a physical device, register for push, and confirm action buttons appear on lock screen
  notifications                                                                                         
  - Tap an action button and confirm invokeServerNotificationAction is called

### DIFF
--- a/iOSNativeSwiftEncryptedNotificationTemplate/EncryptedNotificationTemplate/AppDelegate.swift
+++ b/iOSNativeSwiftEncryptedNotificationTemplate/EncryptedNotificationTemplate/AppDelegate.swift
@@ -32,8 +32,11 @@ class AppDelegate : UIResponder, UIApplicationDelegate {
     
     override init() {
         super.init()
+        // Set the delegate here in init() rather than later in registerForRemotePushNotifications()
+        // to avoid a race condition where a notification could arrive before the delegate is assigned.
+        UNUserNotificationCenter.current().delegate = self
         MobileSyncSDKManager.initializeSDK()
-        
+
         AuthHelper.registerBlock(forCurrentUserChangeNotifications: {
             self.resetViewState {
                 self.setupRootViewController()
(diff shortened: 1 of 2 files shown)


iOSNativeSwiftEncryptedNotificationTemplate/EncryptedNotificationTemplate/AppDelegate.swift
```diff
@@ -42,6 +45,21 @@ class AppDelegate : UIResponder, UIApplicationDelegate {
     }
     
     // MARK: - App delegate lifecycle
+
+    func applicationWillEnterForeground(_ application: UIApplication) {
```
bbirman: I thought the SDK already does the full registration including `PushNotificationManager.sharedInstance().fetchAndStoreNotificationTypes()` by default on foreground ([foreground call to registration ](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/e1b9698fcfa3e4156fbabadcabba973d9c9cdcea/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift#L474) & [notification type call within registration](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/e1b9698fcfa3e4156fbabadcabba973d9c9cdcea/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift#L269)), was there an issue with it?

sfdctaka: This was my oversight. I initially had this while I was debugging the generated app when the notifications weren't coming through. They ended up being just that, notifications not getting sent from the server, and not client having trouble receiving while in background&foreground.

Removed the code.

bbirman (approved):

PR #526, opened 2026-06-01
[auto-generated] Fix CI: Use generic simulator destination in test_template.sh

brandonpage:
## Summary
- Replace hardcoded `name=iPhone 17` simulator destination with portable `generic/platform=iOS Simulator`
- This makes the build step work across all Xcode versions without requiring specific device names

## Root Cause
`test_template.sh` line 139 hardcoded `platform=iOS Simulator,name=iPhone 17` but CI runners (macos-15) use Xcode 16.4 which only ships iPhone 16 simulators. "iPhone 17" requires Xcode 17+.

## Test plan
- [ ] Verify iOSNativeSwiftTemplate builds successfully in CI
- [ ] Verify other iOS templates also build (if they use the same function)
- [ ] Confirm no regression on template generation

### DIFF
--- a/test_template.sh
+++ b/test_template.sh
@@ -136,7 +136,7 @@ build_ios() {
         return 1
     fi
     
-    if xcodebuild $BUILD_TARGET -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 17' build CODE_SIGNING_ALLOWED=NO; then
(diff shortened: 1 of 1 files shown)


wmathurin (approved):

brandonpage (commented):
LGTM

PR #543, opened 2026-07-09
Wire OAuth redirect intent-filter into generated Android manifests (W-23294087)

brandonpage:
## What

Wire the OAuth redirect URI into the generated **Android** app manifest so that, under forced Advanced (browser-based) Authentication (default-ON in Mobile SDK 14.0), the browser redirect is actually delivered back to the app.

Advanced Auth performs login in a Chrome Custom Tab and returns via the app's callback URL. For the OS to deliver that redirect to the app, the SDK `LoginActivity` must declare a redirect `<intent-filter>` whose `<data>` element **mirrors the callback URL** (scheme + host + path). Generated Android apps previously shipped without this filter, so a browser-based login could not complete.

## Changes

- **Native templates** (`AndroidNativeKotlinTemplate`, `AndroidNativeLoginTemplate`, `AndroidIDPTemplate`): activate the SDK `LoginActivity` declaration and add the parameterized redirect `<intent-filter>` with scheme/host/path placeholders. QR-code login guidance preserved.
- **React Native templates** (`ReactNativeTemplate`, `ReactNativeDeferredTemplate`, `ReactNativeTypeScriptTemplate`, `MobileSyncExplorerReactNative`): same manifest change. Also fixed a manifest-merger theme conflict — the SDK `LoginActivity` must use `@style/SalesforceSDK` (matching the SDK library declaration), not `@style/Theme.AppCompat...`, which hard-failed `:app:processDebugMainManifest`.
- **`MobileSyncExplorerKotlinTemplate`**: add the `LoginActivity` declaration (none existed) with the redirect filter.
- **Hybrid templates** (`HybridLocalTemplate`, `HybridRemoteTemplate`): no checked-in manifest, so inject the `LoginActivity` + redirect filter into the Cordova-generated `AndroidManifest.xml` from `template.js prepare()`, guarded on `config.callbackUrlScheme`, idempotent.
- Each `template.js` substitutes the scheme/host/path placeholders from the parsed callback URL.

### `android:path` placeholder ships slash-prefixed

The raw `<data>` line is `android:path="/__INSERT_CALLBACK_URL_PATH_HERE__"`. AAPT rejects any `android:path` that does not begin with `/`, so the **raw** checked-in template must already start with a slash to build (CI builds the template as-is — see Testing). The parsed path (`url.pathname`) already carries its own leading slash, so `template.js` matches the `/`-prefixed token and substitution yields a single-slash path (e.g. `/oauth/done`), never `//oauth/done`. Scheme/host placeholders need no prefix (AAPT accepts them raw).

### Intent-filter design (why mirror the URL, never `*`)

The `<data>` filter mirrors the full callback URL because Android intent-filter matching is a conjunctive subset test — the incoming URI must satisfy **every** attribute the filter declares, so declaring scheme+host+path narrows the filter to exactly this callback. For a **hostless** callback (`scheme:///path`) the host placeholder resolves to the empty string (`android:host=""`), which mirrors the empty authority. It is never `"*"`: `"*"` **is** an Android host wildcard (matches any host, over-broad), whereas `android:host=""` is the correctly-specific mirror.

## Dependency on the Package PR (generation-time, not CI-time)

The scheme/host/path values come from parsed callback-URL config fields (`config.callbackUrlScheme/Host/Path`) populated by the packager. That parsing is added in the companion **SalesforceMobileSDK-Package** PR:

**forcedotcom/SalesforceMobileSDK-Package#341**

**These must ship together** so that apps generated by the CLI get real values substituted into the manifest. Without the Package PR, an app generated by the released CLI would keep the literal placeholder tokens (the app still builds because the path is slash-prefixed — the developer just has to wire the redirect manually).

Note: this repo's CI does **not** exercise that substitution — `test_template.sh` builds the raw checked-in template (`node install.js` → `gradle assembleDebug`) and never runs the packager or `template.js` — so the Package PR is not required for CI here. The dependency is at app-generation time.

## Testing

- **This repo's CI** (`test_template.sh --platform android`) builds each raw checked-in template with `gradle assembleDebug`; the slash-prefixed `android:path` is what lets those raw builds pass AAPT.
- Hybrid generation + build to APK verified via the real CLI path (both hostless and host-bearing callbacks); `aapt2 dump xmltree` confirms the injected filter in the packaged manifest.
- E2E login (real Chrome Custom Tab → redirect returns to app → authenticated) verified via the UITests orchestrator for native_kotlin, hybrid_local, hybrid_remote, and react_native (host-bearing callback shape).

Work item: W-23294087

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/AndroidIDPTemplate/app/src/main/AndroidManifest.xml
+++ b/AndroidIDPTemplate/app/src/main/AndroidManifest.xml
@@ -23,34 +23,34 @@
 
 		<!-- Login activity -->
 		<!--
-            To enable browser based authentication, uncomment the lines below and replace
-            'scheme', 'host' and 'path' with their corresponding values from your connected app.
+            This activity delivers the browser callback Intent to the SDK LoginActivity
+            for browser based (advanced) authentication. The scheme, host and path below
+            are populated from your connected app's callback URL.
 
             For example, if the callback URL of your connected app is
             "testsfdc:///mobilesdk/detect/oauth/done",
-            'scheme' would be "testsfdc", 'host' would be "*" since it doesn't exist, and
-            'path' would be "/mobilesdk/detect/oauth/done".
+            'scheme' would be "testsfdc", 'host' would be empty ("") since it doesn't exist,
+            and 'path' would be "/mobilesdk/detect/oauth/done".
 
             If the callback URL is "sfdc://login.salesforce.com/oauth/done",
             'scheme' would be "sfdc", 'host' would be "login.salesforce.com",
             and 'path' would be "/oauth/done".
         -->
-		<!--
-        <activity android:name="com.salesforce.androidsdk.ui.LoginActivity"
-            android:theme="@style/SalesforceSDK"
-            android:launchMode="singleTask"
-            android:exported="true">
-
-            <intent-filter>
-                <data android:scheme="testsfdc"
-                    android:host="*"
-                    android:path="/mobilesdk/detect/oauth/done" />
-                <action android:name="android.intent.action.VIEW" />
-                <category android:name="android.intent.category.BROWSABLE" />
-                <category android:name="android.intent.category.DEFAULT" />
-            </intent-filter>
-        </activity>
-        -->
+		<activity
+			android:name="com.salesforce.androidsdk.ui.LoginActivity"
+			android:exported="true"
+			android:launchMode="singleTask"
+			android:theme="@style/SalesforceSDK">
+			<intent-filter>
+				<action android:name="android.intent.action.VIEW" />
+				<category android:name="android.intent.category.DEFAULT" />
+				<category android:name="android.intent.category.BROWSABLE" />
+				<data
+					android:scheme="__INSERT_CALLBACK_URL_SCHEME_HERE__"
+					android:host="__INSERT_CALLBACK_URL_HOST_HERE__"
+					android:path="__INSERT_CALLBACK_URL_PATH_HERE__" />
+			</intent-filter>
+		</activity>
 
 	</application>
 </manifest>
--- a/AndroidIDPTemplate/app/src/main/AndroidManifest.xml
+++ b/AndroidIDPTemplate/app/src/main/AndroidManifest.xml
@@ -48,7 +48,7 @@
 				<data
 					android:scheme="__INSERT_CALLBACK_URL_SCHEME_HERE__"
 					android:host="__INSERT_CALLBACK_URL_HOST_HERE__"
-					android:path="__INSERT_CALLBACK_URL_PATH_HERE__" />
+					android:path="/__INSERT_CALLBACK_URL_PATH_HERE__" />
 			</intent-filter>
 		</activity>
 
--- a/AndroidIDPTemplate/template.js
+++ b/AndroidIDPTemplate/template.js
@@ -46,6 +46,7 @@ function prepare(config, replaceInFiles, moveFile, removeFile) {
     var templateStringsXmlFile = path.join('app', 'src', 'main', 'res', 'values', 'strings.xml');
     var templateBootconfigFile = path.join('app', 'src', 'main', 'res', 'values', 'bootconfig.xml');
     var templateServersFile = path.join('app', 'src', 'main', 'res', 'xml', 'servers.xml');
+    var templateAndroidManifestFile = path.join('app', 'src', 'main', 'AndroidManifest.xml');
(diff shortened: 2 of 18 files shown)


wmathurin (commented):
LGTM overall

For hybrid apps, should we tweak the AndroidManifest.xml in postinstall-android.js and the value injection in template.js?

brandonpage:
> LGTM overall
> 
> For hybrid apps, should we tweak the AndroidManifest.xml in postinstall-android.js and the value injection in template.js?

Unless I am mistaken, we can't.  Cordova has not generated the `AndroidManifest.xml` yet when `postinstall-android.js` runs.  I chose to do the work in `template.js prepare()` for all apps.

wmathurin:
> > LGTM overall
> > For hybrid apps, should we tweak the AndroidManifest.xml in postinstall-android.js and the value injection in template.js?
> 
> Unless I am mistaken, we can't. Cordova has not generated the `AndroidManifest.xml` yet when `postinstall-android.js` runs. I chose to do the work in `template.js prepare()` for all apps.

Actually it's already generated - we edit it here: https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/blob/dev/tools/postinstall-android.js#L95

brandonpage:
> > > LGTM overall
> > > For hybrid apps, should we tweak the AndroidManifest.xml in postinstall-android.js and the value injection in template.js?
> > 
> > 
> > Unless I am mistaken, we can't. Cordova has not generated the `AndroidManifest.xml` yet when `postinstall-android.js` runs. I chose to do the work in `template.js prepare()` for all apps.
> 
> Actually it's already generated - we edit it here: https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/blob/dev/tools/postinstall-android.js#L95

You are correct.  I think the code is still correct as is, but I forgot the actual reason 😞.  (That probably warrants a comment)    
  
`postinstall-android.js` runs in the Cordova sub-process and does not have the in memory parsed callback scheme, host and path values.  Seemed much more difficult to get the values there vs using them in `template.js`.

wmathurin:
> > > > LGTM overall
> > > > For hybrid apps, should we tweak the AndroidManifest.xml in postinstall-android.js and the value injection in template.js?
> > > 
> > > 
> > > Unless I am mistaken, we can't. Cordova has not generated the `AndroidManifest.xml` yet when `postinstall-android.js` runs. I chose to do the work in `template.js prepare()` for all apps.
> > 
> > 
> > Actually it's already generated - we edit it here: https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin/blob/dev/tools/postinstall-android.js#L95
> 
> You are correct. I think the code is still correct as is, but I forgot the actual reason 😞. (That probably warrants a comment)
> 
> `postinstall-android.js` runs in the Cordova sub-process and does not have the in memory parsed callback scheme, host and path values. Seemed much more difficult to get the values there vs using them in `template.js`.

I was thinking the block with placeholders could be injected by postinstall-android.js and template.js would just do the replace (like the template.js from android templates do).

That way manually doing `cordova plugin add` would get you a proper app (save for the consumer key/scheme/host etc that need to be manually added in).

For instance test_template.js does `cordova plugin add`.

wmathurin:
As a side note, we should probably update https://git.soma.salesforce.com/SalesforceMobileSDK/SalesforceMobileSDK-Workspace/blob/main/docs/TESTING_CREDENTIALS.md to cover the "manual" injection of scheme/host/path.

brandonpage:
Done — implemented exactly as you suggested. `postinstall-android.js` now injects the `LoginActivity` redirect block with `__INSERT_..._HERE__` placeholder tokens, and `template.js` does the `replace` (mirroring the native/RN templates). So a plain `cordova plugin add` — including `test_template.js` — now gets a proper redirect filter, with the tokens left as a self-documenting breadcrumb for the consumer key / scheme / host / path that must be filled in manually.

The hook change is in CordovaPlugin PR forcedotcom/SalesforceMobileSDK-CordovaPlugin#681. This PR's `template.js` now: if the `LoginActivity` block is present → `replaceInFiles` the three tokens (scheme/host/path); else (older plugin) → inject the block with real values as a fallback. Tokens are byte-identical between the hook and the `template.js` search strings; round-trip verified for hostless / host+path / host-only callbacks with zero leftover tokens.

Also added the comment on `postinstall-android.js` explaining *why* the split exists (the hook runs in the Cordova sub-process without the in-memory parsed callback values).

Good call on `docs/TESTING_CREDENTIALS.md` — tracking that as a follow-up to document the manual scheme/host/path injection for the standalone `cordova plugin add` path.

wmathurin (approved):

PR #546, opened 2026-08-10
fix(security): validate Salesforce Record IDs before SOQL/SmartSQL queries in MobileSyncExplorerSwift (W-23604271)

wmathurin:
## Summary

- Adds `isValidSalesforceId()` — enforces that IDs are exactly 15 or 18 alphanumeric characters before they reach any query
- Guards `fetchContact(id:)` and `localRecord(id:)` in `SObjects.swift` at the query layer (defense-in-depth)
- Guards all three external entry points: notification `targetId` (`NotificationList.swift`), URL scheme deep-link, and `NSUserActivity` handoff (`SceneDelegate.swift`)

Fixes SOQL/SmartSQL injection via unsanitized `targetId` from the Salesforce in-app notification API (W-23604271).

## Test plan

- [ ] Build succeeds with no errors
- [ ] Normal contact tap from notification list opens the contact detail view
- [ ] Normal contact deep-link via URL scheme opens the contact detail view
- [ ] Malformed `targetId` (e.g. containing quotes or SQL keywords) is silently rejected — notification cell rendered as non-tappable, deep-link ignored
- [ ] Locally-created contacts (no Salesforce ID yet) still load correctly via `localRecord(soupID:)` — unaffected by this change

### DIFF
--- a/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
+++ b/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
@@ -31,12 +31,18 @@ import SmartStore
 import MobileSync
 import Combine
 
-// Salesforce Record IDs are exactly 15 or 18 alphanumeric characters.
+// Salesforce Record IDs are exactly 15 or 18 ASCII alphanumeric characters.
 // Enforcing this blocks quote-based injection in SOQL/SmartSQL queries.
 func isValidSalesforceId(_ id: String) -> Bool {
     let len = id.count
     guard len == 15 || len == 18 else { return false }
-    return id.allSatisfy { $0.isLetter || $0.isNumber }
+    return id.allSatisfy { $0.isASCII && ($0.isLetter || $0.isNumber) }
+}
+
+// Soup entry IDs are numeric row identifiers assigned by SmartStore (e.g. "1", "42").
+// Enforcing this blocks injection when a soup entry id is interpolated into SmartSQL.
+func isValidSoupEntryId(_ id: String) -> Bool {
+    return !id.isEmpty && id.allSatisfy { $0 >= "0" && $0 <= "9" }
 }
 
 enum SObjectConstants {
@@ -519,7 +525,7 @@ class SObjectDataManager: ObservableObject {
     }
     
     func localRecord(soupID: String) -> ContactSObjectData? {
-        guard let store = store else {
+        guard let store = store, isValidSoupEntryId(soupID) else {
             return nil
         }
         let queryResult = store.query("select {\(dataSpec.soupName):_soup} from {\(dataSpec.soupName)} where {\(dataSpec.soupName):_soupEntryId} = '\(soupID)'")
--- a/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
+++ b/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
@@ -31,6 +31,14 @@ import SmartStore
 import MobileSync
 import Combine
 
+// Salesforce Record IDs are exactly 15 or 18 alphanumeric characters.
+// Enforcing this blocks quote-based injection in SOQL/SmartSQL queries.
+func isValidSalesforceId(_ id: String) -> Bool {
+    let len = id.count
+    guard len == 15 || len == 18 else { return false }
+    return id.allSatisfy { $0.isLetter || $0.isNumber }
+}
+
 enum SObjectConstants {
     static let kSObjectIdField    = "Id"
 }
@@ -280,6 +288,11 @@ class SObjectDataManager: ObservableObject {
     }
 
     func fetchContact(id: String, completion: @escaping (ContactSObjectData?) -> ()) {
+        guard isValidSalesforceId(id) else {
+            MobileSyncLogger.e(SObjectDataManager.self, message: "fetchContact: invalid id rejected")
+            completion(nil)
+            return
+        }
         let request = RestClient.shared.request(forQuery: "SELECT Id, FirstName, LastName, Title, MobilePhone, Email, Department, HomePhone FROM Contact WHERE Id = '\(id)'", apiVersion: nil)
         RestClient.shared.publisher(for: request)
             .receive(on: RunLoop.main)
@@ -488,7 +501,7 @@ class SObjectDataManager: ObservableObject {
     }
 
     private func localRecord(id: String) -> [String: Any]? {
-        guard let store = store else {
+        guard let store = store, isValidSalesforceId(id) else {
             return nil
         }
         let queryResult = store.query("select {\(dataSpec.soupName):_soup} from {\(dataSpec.soupName)} where {\(dataSpec.soupName):Id} = '\(id)'")
--- a/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
+++ b/MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
@@ -64,6 +64,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             self.window?.rootViewController = UIHostingController(rootView: ContactListView(sObjectManager: sObjectManager, newContact: true))
         } else if let contactRange = url.absoluteString.range(of: "contact/") {
             let id = String(url.absoluteString[contactRange.upperBound...])
+            guard isValidSalesforceId(id) else { return }
             self.window?.rootViewController = UIHostingController(rootView: NavigationStack {
                 ContactDetailView(localId: id, sObjectDataManager: sObjectManager)
             })
@@ -91,7 +92,8 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         let sObjectManager = SObjectDataManager.sharedInstance(for: UserAccountManager.shared.currentUserAccount!)
 
         if let userActivity = userActivity, userActivity.title == openDetailPath,
-           let selectionId = userActivity.userInfo?[openDetailRecordIdKey] as? String {
+           let selectionId = userActivity.userInfo?[openDetailRecordIdKey] as? String,
+           isValidSalesforceId(selectionId) {
             self.window?.rootViewController = UIHostingController(rootView:  NavigationStack {
                 ContactDetailView(localId: selectionId, sObjectDataManager: sObjectManager)
             })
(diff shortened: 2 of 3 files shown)


MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
```diff
@@ -91,7 +92,8 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         let sObjectManager = SObjectDataManager.sharedInstance(for: UserAccountManager.shared.currentUserAccount!)
 
         if let userActivity = userActivity, userActivity.title == openDetailPath,
-           let selectionId = userActivity.userInfo?[openDetailRecordIdKey] as? String {
+           let selectionId = userActivity.userInfo?[openDetailRecordIdKey] as? String,
+           isValidSalesforceId(selectionId) {
```
JohnsonEricAtSalesforce: This guards the `NSUserActivity` handoff path, but `selectionId` here is the value produced by `ContactListViewModel.itemProvider`, which is `contact.id` = `_soupEntryId` (a numeric soup row id from `Contacts.swift:146`), and it's consumed by `localRecord(soupID:)` which queries `_soupEntryId`. `isValidSalesforceId` requires 15/18 chars, so a real soup entry id (1–3 digits) fails the check and the detail restore falls through to the `else` (`Tabs`) branch. Before this change the path had no validation and worked. Consider a soup-entry-id-appropriate check here (e.g. numeric) and reserve `isValidSalesforceId` for the Salesforce-Id path.

MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
```diff
@@ -64,6 +64,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             self.window?.rootViewController = UIHostingController(rootView: ContactListView(sObjectManager: sObjectManager, newContact: true))
         } else if let contactRange = url.absoluteString.range(of: "contact/") {
             let id = String(url.absoluteString[contactRange.upperBound...])
+            guard isValidSalesforceId(id) else { return }
```
JohnsonEricAtSalesforce: Same mismatch as the handoff path. The `contact/<id>` deep-link is produced by the RecentContacts widget (`RecentContactsWidget.swift:78`): `URL(string: "mobilesyncexplorerswift://contact/\(contact.id)")`, where `contact` is a `ContactSummary` whose `id` is the numeric `_soupEntryId` (populated in `ContactListViewModel.addRecentContact` at line 118 from `contact.id`). The widget's placeholder data uses ids `"1"`, `"2"`, `"3"`, so the app emits URLs like `contact/1`. Here that value routes to `ContactDetailView(localId:)` → `localRecord(soupID:)`, which queries `_soupEntryId` — so a valid id is 1–3 digits, and `isValidSalesforceId` (15/18 chars) rejects it. Consider a soup-entry-id-appropriate check here as well. (If deep-links were instead intended to carry a Salesforce Id, note the query already targets `_soupEntryId` — that mismatch predates this PR and would need its own decision.)

MobileSyncExplorerSwift/MobileSyncExplorerSwift/UI/Notifications/NotificationList.swift
```diff
@@ -54,7 +54,7 @@ struct ListView: View {
         List {
             ForEach(model.notifications, id: \.id) { notification in
                 VStack {
-                    if notification.targetId.starts(with: "003") { // Only contacts are tappable
+                    if notification.targetId.starts(with: "003") && isValidSalesforceId(notification.targetId) { // Only contacts are tappable
```
JohnsonEricAtSalesforce, line 57: This is the entry point where `isValidSalesforceId` is the right check: `notification.targetId` is a real Salesforce Id and routes to `fetchContact(id:)` → `SELECT … WHERE Id = '…'`. The added guard here reads well alongside the existing `starts(with: "003")` contact check.

MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
```diff
@@ -31,6 +31,14 @@ import SmartStore
 import MobileSync
 import Combine
 
+// Salesforce Record IDs are exactly 15 or 18 alphanumeric characters.
+// Enforcing this blocks quote-based injection in SOQL/SmartSQL queries.
+func isValidSalesforceId(_ id: String) -> Bool {
+    let len = id.count
+    guard len == 15 || len == 18 else { return false }
+    return id.allSatisfy { $0.isLetter || $0.isNumber }
```
JohnsonEricAtSalesforce: `$0.isLetter || $0.isNumber` accepts Unicode letters/digits (accented letters, non-ASCII digits), whereas Salesforce Ids are ASCII `[a-zA-Z0-9]`. It still blocks quote-based injection (a quote is neither letter nor number), so this isn't a security gap — but it's looser than the SDK-side twins (Android 2992 / iOS 4131), which escape against ASCII. If you'd like the sample to model the stricter contract, an explicit ASCII-alphanumeric check would match. Minor.

JohnsonEricAtSalesforce (changes requested):
Thanks for hardening the sample against SOQL/SmartSQL injection — validating record IDs at the untrusted notification/deep-link boundary is the right instinct, and the injection goal is met: quote/keyword payloads fail `isValidSalesforceId` before reaching any query on all three entry points.

I traced each user-influenced ID from its entry point to its SQL sink, and I believe two of the three guarded paths have a functional regression: they validate the wrong ID format and will reject legitimate input.

**The routing (at `e1b4a57`):**

| Entry point | View init | Query method | Field queried | ID value carried |
|---|---|---|---|---|
| Notification tap | `ContactDetailView(id:)` | `fetchContact(id:)` | SOQL `Id = '…'` | real Salesforce Id (`003…`) |
| URL deep-link | `ContactDetailView(localId:)` | `localRecord(soupID:)` | SmartSQL `_soupEntryId = '…'` | numeric soup entry id |
| Handoff (`NSUserActivity`) | `ContactDetailView(localId:)` | `localRecord(soupID:)` | SmartSQL `_soupEntryId = '…'` | numeric soup entry id |

`ContactSObjectData.id` (`Contacts.swift:146`) returns `_soupEntryId` as a string — a numeric SQLite row id (e.g. `"5"`, `"42"`), not a Salesforce Id. The handoff producer (`ContactListViewModel.itemProvider`) puts that numeric id into the activity, and the consumer queries `_soupEntryId` with it. So on the deep-link and handoff paths, `isValidSalesforceId` (which requires length 15 or 18) rejects every realistic value, and the detail view silently fails to open (handoff falls through to the plain `Tabs` view). The notification path is the one place the Salesforce-Id check is correct.

This compiles cleanly, so it isn't caught by CI — the template job is build-only (`test_template.sh` runs `xcodebuild … build`; TESTING.md notes it doesn't run the app). All checks are green at `e1b4a57`, which is expected and doesn't speak to this runtime behavior.

Suggested direction (your call on the shape): keep `isValidSalesforceId` on the `fetchContact` / Salesforce-Id path, and validate the `_soupEntryId` paths with a check appropriate to a soup entry id (e.g. numeric). If instead deep-links/handoff are meant to carry Salesforce Ids, then the pre-existing `_soupEntryId` query on those paths is already querying the wrong field and it's worth settling that first.

Two smaller points are in the inline comments.

MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
wmathurin: {'reply': 'Fixed. Added `isValidSoupEntryId()` in `SObjects.swift` (accepts non-empty all-digit strings) and switched both the NSUserActivity handoff path and `localRecord(soupID:)` to use it instead of `isValidSalesforceId`. A soup entry id like `"42"` now passes where the 15/18-char Salesforce Id check was incorrectly rejecting it.', 'resolve': True}

MobileSyncExplorerSwift/MobileSyncExplorerSwift/SceneDelegate.swift
wmathurin: {'reply': 'Fixed. The `contact/<id>` deep-link path now uses `isValidSoupEntryId` (numeric check) rather than `isValidSalesforceId`. The `contact/<id>` URL is produced by the RecentContacts widget with `_soupEntryId` values, so the numeric guard is the right fit here.', 'resolve': True}

MobileSyncExplorerSwift/MobileSyncExplorerSwift/UI/Notifications/NotificationList.swift
wmathurin: {'reply': 'Agreed — `isValidSalesforceId` is the correct check on this path since `notification.targetId` is a real Salesforce Id routing to `fetchContact(id:)`. No change needed here.', 'resolve': True}

MobileSyncExplorerSwift/MobileSyncExplorerSwift/SObjects/SObjects.swift
wmathurin: {'reply': 'Fixed. `isValidSalesforceId` now uses `$0.isASCII && ($0.isLetter || $0.isNumber)` to restrict to ASCII `[a-zA-Z0-9]`, matching the stricter contract from the SDK-side twins (Android 2992 / iOS 4131).', 'resolve': True}

JohnsonEricAtSalesforce (approved):
Thanks for the quick turnaround — the fix at `93fa4dc` addresses all four points. I re-traced every caller of the `_soupEntryId` path and confirmed:

- The deep-link and handoff guards now use `isValidSoupEntryId`, so legitimate numeric soup entry ids (e.g. the `contact/1` URLs the RecentContacts widget emits) are accepted again, while injection payloads are still rejected.
- Every value reaching `localRecord(soupID:)` and `ContactDetailView(localId:)` is a numeric `_soupEntryId` (`contact.id`), so the numeric guard rejects nothing legitimate.
- `isValidSalesforceId` restricting to ASCII matches the SDK-side contract.
- Adding the `isValidSoupEntryId` guard inside `localRecord(soupID:)` is a nice defense-in-depth touch at the sink itself.

Injection is blocked at all entry points, and the earlier functional regression on the local-navigation paths is resolved. Approving. (CI is green at `93fa4dc`; note the template job is build-only, so the runtime behavior above was verified by code trace rather than an executed test.)

PR #547, opened 2026-08-22
[Templates][Android] presentOptInDialog call fails to compile — method removed from BiometricAuthenticationManager

JohnsonEricAtSalesforce:
## Summary

`AndroidNativeLoginTemplate` fails to compile because `BiometricAuthenticationManager.presentOptInDialog(FragmentManager)` was removed upstream in `SalesforceMobileSDK-Android` — the SDK now auto-presents the biometric opt-in dialog during login instead, so no replacement call is needed.

- Removed the `onPostResume()` override in `MainActivity.kt` that called it, plus its now-unused imports (including a stray `android.service.autofill.Validators.or` that was shadowing Kotlin's `Int.or`)
- Confirmed via repo-wide grep this is the only affected template; `NativeLogin.kt`'s biometric code uses the separate, unaffected `presentBiometricAuth` login/unlock API

**Note for reviewers:** a generated build of this template may still fail on an unrelated, separately-tracked `logout()` signature change in the same file — not a regression from this PR.

## Test plan

- [x] `./gradlew assembleDebug` succeeds locally against current `SalesforceMobileSDK-Android` `dev`
- [ ] Automatic opt-in dialog visually confirmed on a biometric-capable device (not available in this environment; confirmed via SDK source that the flow needs no app-side hook)

### DIFF
--- a/AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate/MainActivity.kt
+++ b/AndroidNativeLoginTemplate/app/src/main/java/com/salesforce/androidnativelogintemplate/MainActivity.kt
@@ -26,13 +26,9 @@
  */
 package com.salesforce.androidnativelogintemplate
 
-import androidx.biometric.BiometricManager
-import androidx.biometric.BiometricManager.Authenticators.BIOMETRIC_STRONG
-import androidx.biometric.BiometricManager.Authenticators.BIOMETRIC_WEAK
 import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES.UPSIDE_DOWN_CAKE
 import android.os.Bundle
-import android.service.autofill.Validators.or
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ArrayAdapter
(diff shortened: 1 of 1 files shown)


JohnsonEricAtSalesforce:
\`AndroidNativeLoginTemplate\`'s CI job on this PR fails on an unrelated, separately-tracked issue — not a regression introduced here. This repo currently has two independent fixes in flight for \`AndroidNativeLoginTemplate\`, each addressing one of two distinct compile breaks in the same file caused by separate upstream \`SalesforceMobileSDK-Android\` changes:

- #547 (this PR) — removes the stale \`presentOptInDialog()\` call (method removed from \`BiometricAuthenticationManager\`)
- #548 — fixes \`logout(this)\` call sites broken by \`SalesforceSDKManager.logout()\`'s parameter reorder

Each PR's branch is missing the other's fix, so \`AndroidNativeLoginTemplate\` will fail to compile on **either PR individually** until both merge — this is expected, not a sign either fix is wrong. Verified locally: applying both fixes together compiles clean. CI on \`dev\` will go green once both merge; no further action needed on either PR to resolve this specific failure.

JohnsonEricAtSalesforce:
**CI status: expected failure, not a regression**

`android-pr / test-android (AndroidNativeLoginTemplate)` fails in this PR's CI run, but not on the code this PR touches. The failure is a Kotlin compile error at `MainActivity.kt:112`:

```
Argument type mismatch: actual type is 'MainActivity', but 'Account?' was expected.
No value passed for parameter 'frontActivity'.
```

That line is `SalesforceSDKManager.getInstance().logout(this)` — unrelated to the `presentOptInDialog` removal in this PR. It's failing because `SalesforceMobileSDK-Android`'s `logout()` signature changed upstream (separately tracked, fix in flight on another branch/PR). This template file happens to be affected by both SDK changes at once, so a full green build here is blocked on that other fix landing first.

Confirmed locally that this PR's fix is what it claims to be: with only the `presentOptInDialog` removal applied, the build gets past that line cleanly and fails only at the unrelated `logout()` call — matching this CI run exactly.

**Path forward:** this PR is ready on its own merits. Full CI green on this template requires the companion `logout()` fix to merge as well (either before or after this one — order doesn't matter, they're independent changes to the same file). No further action needed here unless a reviewer wants to hold merge until both land together.

wmathurin (approved):

brandonpage (approved):
👍 The prompt is now automatic on the first login.

PR #548, opened 2026-08-22
[Templates][Android] logout(this) call sites fail to compile against SalesforceSDKManager.logout() param reorder

JohnsonEricAtSalesforce:
## Summary
`SalesforceSDKManager.logout()` reordered its parameters (`account` now precedes `frontActivity`), breaking positional `logout(this)` calls in three Android templates. Fixes each site with the named-argument form, and — per internal reviewer feedback — also passes an explicit `reason = OAuth2.LogoutReason.USER_LOGOUT` at the three call sites that fire from an actual user tap, matching the SDK's own internal convention (e.g. `SalesforceSDKManager`'s dev-menu Logout action). The one automatic/no-current-user cleanup call in `ContactsActivity` is left at the default reason since it isn't user-initiated.

## Changes
- `AndroidNativeKotlinTemplate/.../MainActivity.kt`
- `AndroidNativeLoginTemplate/.../MainActivity.kt`
- `MobileSyncExplorerKotlinTemplate/.../ContactsActivity.kt` (2 call sites; only the user-click one gets `reason`)

## Test plan
- [x] `./gradlew :app:compileDebugKotlin` locally against the reordered SDK signature — `AndroidNativeKotlinTemplate` and `MobileSyncExplorerKotlinTemplate` build clean
- [x] `AndroidNativeLoginTemplate` confirmed to compile clean on this fix (isolated from an unrelated, separately-tracked `presentOptInDialog` break in the same file)
- [x] End-to-end integration re-test via the Package repo's CLI harness — passing

### DIFF
--- a/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
+++ b/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/MainActivity.kt
@@ -137,7 +137,7 @@ class MainActivity : SalesforceActivity() {
      */
     @Suppress("UNUSED", "UNUSED_PARAMETER")
     fun onLogoutClick(v: View) {
-        SalesforceSDKManager.getInstance().logout(this)
+        SalesforceSDKManager.getInstance().logout(frontActivity = this)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


JohnsonEricAtSalesforce:
\`AndroidNativeLoginTemplate\`'s CI job on this PR fails on an unrelated, separately-tracked issue — not a regression introduced here. This repo currently has two independent fixes in flight for \`AndroidNativeLoginTemplate\`, each addressing one of two distinct compile breaks in the same file caused by separate upstream \`SalesforceMobileSDK-Android\` changes:

- #547 — removes the stale \`presentOptInDialog()\` call (method removed from \`BiometricAuthenticationManager\`)
- #548 (this PR) — fixes \`logout(this)\` call sites broken by \`SalesforceSDKManager.logout()\`'s parameter reorder

Each PR's branch is missing the other's fix, so \`AndroidNativeLoginTemplate\` will fail to compile on **either PR individually** until both merge — this is expected, not a sign either fix is wrong. Verified locally: applying both fixes together compiles clean. CI on \`dev\` will go green once both merge; no further action needed on either PR to resolve this specific failure.

wmathurin (approved):

brandonpage (approved):

PR #554, opened 2026-09-02
[iOS] Native Login template: authenticated Accounts screen renders blank after login/registration until app is backgrounded and foregrounded

JohnsonEricAtSalesforce:
## Problem

After a successful login or registration in the iOS Native Login template, the authenticated
Accounts screen renders blank. The screen only appears correctly after backgrounding and
foregrounding the app.

## Root cause

`SceneDelegate.setupRootViewController()` assigns a new `rootViewController` to the window but
never calls `window.makeKeyAndVisible()` afterward, so the new view hierarchy is never forced to
repaint. Backgrounding/foregrounding incidentally triggers a repaint through a different code
path, masking the bug.

## Fix

Call `window?.makeKeyAndVisible()` immediately after assigning the authenticated root view
controller in `setupRootViewController()`.

## Related fix found during testing

Testing also surfaced a crash in the template's out-of-the-box default configuration (reCAPTCHA/
password-less login left commented out, as the template ships): `NativeLoginViewFactory`'s
no-argument `create()` builds `NativeLoginView` without injecting a `ReCaptchaClientObservable`
into the environment, so any access to the view's `@EnvironmentObject` fatals immediately on a tap
of Register, Reset Password, or Passwordless-OTP Request. This does not affect the
manually-configured reCAPTCHA path, which has always injected the observable correctly.

Fixed by always injecting a `ReCaptchaClientObservable` — configured when reCAPTCHA is set up,
and a no-op placeholder otherwise. Since that placeholder's `reCaptchaClient` is legitimately
unconfigured, the three OTP-request handlers (Register, Reset Password, Passwordless-OTP Request)
now guard on a configured `reCaptchaClient` before proceeding, showing a clear error message
instead of silently no-oping through optional chaining — which would otherwise leave the UI stuck
on a progress spinner indefinitely with no error and no crash.

## Test plan

- [x] Reproduced the original blank-screen bug via the plain username/password Login flow
      (no reCAPTCHA configuration needed for this path)
- [x] Verified the fix resolves it: Accounts screen renders immediately after login, without
      backgrounding/foregrounding
- [x] Reproduced the crash on Register / Reset Password / Passwordless-OTP request in an
      out-of-the-box generated app; verified it no longer crashes after the fix
- [x] Verified the unconfigured-reCAPTCHA path now shows a clear error instead of hanging
- [x] Verified the plain Login flow (no reCAPTCHA required) is unaffected by any of the above
      changes
- [x] `xcodebuild build` succeeds for `iOSNativeLoginTemplate` with no new warnings
- [x] `test_template.sh --template iOSNativeLoginTemplate --platform ios` passes

### DIFF
--- a/iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
+++ b/iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
@@ -435,7 +435,12 @@ struct NativeLoginView: View {
             
             // Guard for the new reCAPTCHA token.
             guard let reCaptchaToken = reCaptchaExecuteResult else {
-                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA signup action token due to error with description '\(error?.localizedDescription ?? "(A description wasn't provided.)")'.")
+                let description = error?.localizedDescription ?? "(A description wasn't provided.)"
+                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA signup action token due to error with description '\(description)'.")
+
+                // Clear the progress indicator and inform the user.
+                isAuthenticating = false
+                errorMessage("Could not obtain a reCAPTCHA token due to an error with description '\(description)'.")
                 return
             }
             
@@ -518,7 +523,12 @@ struct NativeLoginView: View {
             
             // Guard for the new reCAPTCHA token.
             guard let reCaptchaToken = reCaptchaExecuteResult else {
-                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA forgot password action token due to error with description '\(error?.localizedDescription ?? "(A description wasn't provided.)")'.")
+                let description = error?.localizedDescription ?? "(A description wasn't provided.)"
+                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA forgot password action token due to error with description '\(description)'.")
+
+                // Clear the progress indicator and inform the user.
+                isAuthenticating = false
+                errorMessage("Could not obtain a reCAPTCHA token due to an error with description '\(description)'.")
                 return
             }
             
@@ -593,7 +603,12 @@ struct NativeLoginView: View {
             
             // Guard for the new reCAPTCHA token.
             guard let reCaptchaToken = reCaptchaExecuteResult else {
-                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA login action token due to error with description '\(error?.localizedDescription ?? "(A description wasn't provided.)")'.")
+                let description = error?.localizedDescription ?? "(A description wasn't provided.)"
+                SalesforceLogger.e(AppDelegate.self, message: "Could not obtain a reCAPTCHA login action token due to error with description '\(description)'.")
+
+                // Clear the progress indicator and inform the user.
+                isAuthenticating = false
+                errorMessage("Could not obtain a reCAPTCHA token due to an error with description '\(description)'.")
                 return
             }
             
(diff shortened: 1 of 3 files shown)


iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -420,12 +420,18 @@ struct NativeLoginView: View {
     private func onRequestOtpForRegistrationTapped() {
         // Reset the message.
         messageReset()
-        
+
+        // Guard for a configured reCAPTCHA client.
+        guard let reCaptchaClient = reCaptchaClientObservable.reCaptchaClient else {
```
brandonpage: `reCaptchaClient` is also `nil` while `ReCaptchaClientObservable(reCaptchaSiteKey:)` is asynchronously calling `Recaptcha.getClient`, and it remains nil if initialization fails. A correctly configured app can therefore show “reCAPTCHA is not configured” when the user acts before initialization finishes or after an initialization error. Could the observable expose explicit disabled/loading/ready/failed states, then disable or wait while loading and surface the actual initialization error? The same issue applies to the guards in the reset-password and passwordless-login handlers.

iOSNativeLoginTemplate/iOSNativeLoginTemplate/NativeLoginView.swift
```diff
@@ -420,12 +420,18 @@ struct NativeLoginView: View {
...
+            return
+        }
+
         // Show the progress indicator.
         isAuthenticating = true
-        
+
         // Execute for a new reCAPTCHA token.
-        reCaptchaClientObservable.reCaptchaClient?.execute(
+        reCaptchaClient.execute(
```
brandonpage, line 432: After `isAuthenticating` becomes true, a failed reCAPTCHA `execute` only logs and returns. The view therefore remains blurred with its progress indicator indefinitely. Please reset `isAuthenticating` and show an error on this path, including the equivalent branches in the reset-password and passwordless-login handlers.

JohnsonEricAtSalesforce:
Thanks for the thorough review, @brandonpage — both points are addressed:

**On the reCAPTCHA state conflation:** `ReCaptchaClientObservable` now exposes an explicit `ReCaptchaClientState` enum (`disabled` / `loading` / `ready` / `failed`) instead of a bare optional client. `NativeLoginView`'s three OTP-request handlers now call a `readyReCaptchaClient()` helper that switches on this state and shows the correct message for each case — including surfacing the actual initialization error on `.failed`, as you suggested. Applied identically to registration, reset-password, and passwordless-login.
Commit: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/554/commits/ed9e134

**On the execute-failure hang:** all three `execute(...)` completion closures now reset `isAuthenticating = false` and show an error message on the failure path, instead of only logging and returning. The view no longer stays blurred with a spinning indicator indefinitely.
Commit: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/554/commits/e837566

Both changes were verified with a local build and the repo's template test harness. Let us know if anything still looks off.

JohnsonEricAtSalesforce:
While addressing the reCAPTCHA state-machine feedback above, we noticed that `ReCaptchaClientState.failed` is a terminal state: once client initialization fails, the observable never retries, so an affected app stays in that failed state (with password-less login unavailable) until relaunched.

We're deferring adding retry logic for now — this is a minimal starter template, not a production reCAPTCHA integration, so we kept the failure path to logging + surfacing the error via the existing state machine rather than adding backoff/retry behavior. We added a comment at the failure site in `SceneDelegate.swift` pointing developers building on this template at the right place to add their own retry logic if they need it.

Commit: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/554/commits/e49ab6e

brandonpage (approved):

PR #555, opened 2026-09-08
[MSDK Templates] AndroidNativeKotlinTemplate Doesn't Display Actionable Notifications For API 32

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🥁
  While testing, I noticed that the Actionable Notifications support in the Android Native Kotlin Template was silently dropping notifications on API 32 client-side.  Here's a really simple fix I tested today on API 32 that resolves that issue.

### DIFF
--- a/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/PushNotificationsAdapter.kt
+++ b/AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/PushNotificationsAdapter.kt
@@ -31,6 +31,8 @@ import android.app.PendingIntent.FLAG_IMMUTABLE
 import android.app.PendingIntent.getBroadcast
 import android.content.Intent
 import android.content.pm.PackageManager.PERMISSION_GRANTED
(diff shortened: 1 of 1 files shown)


AndroidNativeKotlinTemplate/app/src/main/java/com/salesforce/androidnativekotlintemplate/PushNotificationsAdapter.kt
```diff
@@ -150,7 +152,7 @@ internal class PushNotificationsAdapter : PushNotificationInterface {
         }.also { builder ->
 
             // Build and display the notification, if possible.
-            if (context.checkSelfPermission(POST_NOTIFICATIONS) == PERMISSION_GRANTED) {
+            if (SDK_INT < TIRAMISU || context.checkSelfPermission(POST_NOTIFICATIONS) == PERMISSION_GRANTED) {
```
JohnsonEricAtSalesforce, line 155: This maps to the documentation and API Level support for POST_NOTIFICATIONS, which I'd simply missed with creating the original template.

wmathurin (approved):

